PR #11, opened 2015-01-02
Check assert statements with nonsense expressions

bitglue:
(authored by @clintonroy, patch submitted via LaunchPad)

Fixes lp:1405778.

I've come across numerous accidental uses of assert where the test
expression is a tuple of the actual test expression and the message,
which always succeeds as the bool of a non empty tuple is always true.

This patch catches the most obvious constants boolean expressions assed
along, certainly not all of them, but there shouldn't be any false
positives.

### DIFF
--- a/pyflakes/test/test_asserts.py
+++ b/pyflakes/test/test_asserts.py
@@ -1,0 +1,39 @@
+from pyflakes import messages as m
+from pyflakes.test.harness import TestCase
+
+class Test(TestCase):
+    def test_Asserts_intentionally_simple_pass(self):
+        self.flakes("""
+        assert True
+        assert False
+        """)
+
+    def test_Asserts_nontrivial_pass(self):
+        self.flakes("""
+        d = {'a': 1, 'b': 2}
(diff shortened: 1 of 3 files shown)


pyflakes/test/test_asserts.py
```diff
@@ -0,0 +1,39 @@
...
+        n = 10
+        s = {1, 2}
+        t = (1, 2)
+        needle = 'needle'
+        assert d == {'a': 3}
+        assert n == 10
+        assert l == [1, 2]
+        assert s == {1, 2}
+        assert t == (1, 2)
+        assert needle in "haystack"
```
sigmavirus24: Can you add a test for something like

```python
assert conditional, "Message in case of assertion error"
```

That would be an implicit tuple if the assert statement didn't handle it differently (and tuples are tested below)

bitglue:
@clintonroy, I'm closing this because it can not be accepted with the test failures as the are. Please feel free to submit a new PR with a revised patch if you want to tackle those issues.

PR #12, opened 2015-02-14
[#885140] Update del handling in if and while statements.

adiroiban:
Problem
=======

"del" handling in conditional blocks is to optimistic and assumes that the branch will always be executed.

"del" handling does not take into consideration the fact that a variable is used as part of a while test.

Changes
=======

Update handleNodeDelete to not remove the variable from the scope when del call is on a conditional branch or when is part of while's test scope.

I tried to keep the tests simple. In case there are any problems I will update the test suite and update the code.

I wrote testes based on examples from https://bugs.launchpad.net/pyflakes/+bug/885140 and this code from Twisted https://github.com/twisted/twisted/blob/trunk/twisted/python/threadpool.py#L190

Tests
====

```$ python setup.py test -q``` all tests passed.

I also run tox, but I only have python 2.7 and 3.4 installed.

I have tested the new code on Twisted trunk and on my private project it looks good.
No false positives....but since the code was already clean I am not sure if this change introduces any false negatives.

Please take a look and let me know what needs to be changes.

Thanks!

### DIFF
--- a/pyflakes/checker.py
+++ b/pyflakes/checker.py
@@ -531,9 +531,47 @@ class Checker(object):
         self.addBinding(node, binding)
 
     def handleNodeDelete(self, node):
+
+        def on_conditional_branch():
+            """
+            Return `True` if node is part of a conditional branch.
+            """
+            current = getattr(node, 'parent', None)
+            while current:
+                if isinstance(current, ast.If):
+                    return True
+                current = getattr(current, 'parent', None)
+            return False
+
+        def is_part_of_while_conditional():
+            """
+            Return `True` if node is part of left or right side of a while
+            test.
+            """
+            current = getattr(node, 'parent', None)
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by bitglue** [comment]
Looks pretty good. Thanks for the PR.

I wonder if it might make more sense and be simpler to ignore `del` unconditionally in a `while` block. Consider:

```python
def test():
    foo = "bar"
    while False:
        del foo
    print foo
```

I was also a bit confused by the notion of "left side" and "right side" of a `while` statement:

image

(sorry couldn't resist)

adiroiban:
True, while body is also conditionally called.
I left the previous while test there just in case this will be refactored.

Please take a look at latest changes.

Thanks!

bitglue:
Looks great, thanks.

PR #13, opened 2015-02-14
[#1332451] Add an error for return outside of function.

adiroiban:
Problem
======
See: https://bugs.launchpad.net/pyflakes/+bug/1332451

The code on master was already ignoring this error and no longer crashing.

Changes
=======

Create a dedicated errror but not sure if this is needed as it is already raised by python as a syntax error.

Check that changes make sense of just close this pull request and the Launchpad ticket.

Thanks!


bitglue:
Looks good, thanks.

The reason these errors aren't already caught is that pyflakes parses, but does not compile the code. See  #8.

PR #14, opened 2015-03-15
Properly support list comprehensions in Python 3

myint:
I've modified the code to match the new behavior in Python 3.

https://docs.python.org/3/reference/expressions.html#displays-for-lists-sets-and-dictionaries

> Note that the comprehension is executed in a separate scope, so names assigned to in the target list don’t “leak” into the enclosing scope.

### DIFF
--- a/pyflakes/checker.py
+++ b/pyflakes/checker.py
@@ -672,7 +672,7 @@ class Checker(object):
         EQ = NOTEQ = LT = LTE = GT = GTE = IS = ISNOT = IN = NOTIN = ignore
 
     # additional node types
-    LISTCOMP = COMPREHENSION = KEYWORD = handleChildren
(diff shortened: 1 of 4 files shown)


bitglue:
Thanks for the PR. I'll give this a look tonight.

bitglue:
Code looks good, but it's a little unclear from the context in the code what "leaks" means. The explanation in the PR description here  makes sense though, so can we incorporate that into the code?

In particular, just "no longer leaks" as in the conditional tests of the skip had me wondering "leaks what?" Maybe something a little more verbose, like "in python3 list comprehensions execute in a separate scope" would be better.

myint:
Sure. I've added the more verbose message.

bitglue:
Looks good, thanks. I'll merge once Travis gives it a :white_check_mark:

PR #16, opened 2015-04-04
Windows fix

osuchw:
It is AttributeError that is thrown on Windows

### DIFF
--- a/pyflakes/api.py
+++ b/pyflakes/api.py
@@ -139,7 +139,7 @@ def main(prog=None):
     try:
         signal.signal(signal.SIGINT, lambda sig, f: sys.exit('... stopped'))
         signal.signal(signal.SIGPIPE, lambda sig, f: sys.exit(1))
-    except ValueError:
+    except AttributeError:
         pass    # SIGPIPE is not supported on Windows
 
     parser = optparse.OptionParser(prog=prog, version=__version__)


bitglue:
I suspect this is only true for the particular version of Windows or Python that you are running. For example, see

https://docs.python.org/2/library/signal.html#signal.signal

> On Windows, signal() can only be called with SIGABRT, SIGFPE, SIGILL, SIGINT, SIGSEGV, or SIGTERM. A ValueError will be raised in any other case.

Since we don't have any automated tests on Windows, can you provide any further documentation to back up this change?

myint:
@bitglue, I think the documentation is incorrect. This has come up in the past before.

https://github.com/jcrocholl/pep8/pull/280
https://bugs.python.org/issue21382

bitglue:
Well, the Python bug is somewhat insightful:

> Because the C code defines module-level constants on the basis of whether the platform C-lib defines them, different platforms will have different module attributes.

So it seems there are two problems one might encounter. Firstly, `os.SIGPIPE` might not be defined. Then you get `AttributeError`. Or, it might be defined, but then ultimately invalid. Then you get `ValueError`.

I don't really know much about Python on Windows, but if it's anything like I imagine, there are many ways to do it, and I imagine in at least some of them, either of these situations might occur, so we should probably anticipate each.

However, I think broadly catching either exception is too broad, and could potentially swallow errors that shouldn't be swallowed. Perhaps the exception handling can be more narrowly targeted at the case of `signal.SIGPIPE` being undefined?

bitglue:
@osuchw can you see if #18 addresses your problem?

osuchw:
Yes it does fix the failing import.  Thank you.

bitglue:
superseded by #18.

PR #19, opened 2015-04-26
Add names declared with 'global' statement to global scope.

ryneeverett:


bitglue:
hmm...what about this case?

```python
def c(): bar
def d(): global bar; bar = 1
```

ryneeverett:
I don't think github gives notifications of additional commits to PR's, so <ping>.

AlexChesters:
+1

bitglue:
OK, looking pretty good, but there are a few more cases:

```python
import foo
def f():
  global foo
def g():
  foo.is_used()
```

This will report that `foo` is unused, but it is used (in `g()`).

Also, though the code seems to handle multiple names in one global statement (`global foo, bar`), there seems to be no test for such a case.

ryneeverett:
I think `global` statements should never trigger `UnusedImport` errors

- When used in module scope `global` statements are ignored.
- Pyflakes is ignorant of whether the function containing a `global` statement is called before or after the import it shadows is used.

bitglue:
"never" seems to be too much. Consider in this case, there really is no way the import could ever be used:

```python
import fu
def f(): global fu
```

The `global` really is irrelevant: there's no way that program, with the functions called in any order, could possibly use the import. I think there should be an error in this case, and the test should remain.

But this case is different:

```python
import foo
def f():
  global foo
def g():
  foo.is_used()
```

Here, `foo` certainly is used. The `global foo` isn't really relevant, anyway, since even though it declares `foo` to be a global, it doesn't use it or change it any way. So there should not be an error in this case.

For example, consider:

```python
import time
def f():
    global time
    print time

f()
```

For me, that prints `<module 'time' (built-in)>`, just as it would if `global time` were not in the program.

Is there a way to accommodate both these cases?

ryneeverett:
You're right, my thinking was muddled. I'll reformulate: *`global` statements never have an effect on whether or not there is an unused import*.

Whether they *trigger* an `UnusedImport` message is a contingent implementation detail, which I hope I've successfully avoided.

Check out the current branch. I did some naughty history revision so the last couple commits are quite different.

bitglue:
This looks good. I fall into the history-revision-is-not-naughty-if-it's-not-yet-merged camp, so could you please squash and rebase all these commits into one commit, with a well-written commit message and summary? Then I'll merge.

Thanks for the fix, and sticking with it. Really appreciate it.

ryneeverett:
Done. Hope that's a decent summary.

And thank you for a reasonable linter. I've spent more time just trying to figure out how to configure *other-linter-that-will-not-be-named* than I did making this PR.

sigmavirus24:
@ryneeverett you mean the linter that generates a ~100 line config file by default for you?

ryneeverett:
@sigmavirus24 I have a lot of respect for the people involved in that project, but yes, that would be the one.

sigmavirus24:
@ryneeverett I do too. It's a good project but I think that's why landscape.io is necessary. They have a really good default config for that project and use pyflakes as well

ryneeverett:
@sigmavirus24 Tools that help with code review are great but it's even better if you get the feedback before committing.

It would be cool if there were a tool that could analyze an existing code base and spit out a pylint (oops) config file that respects existing conventions. It could perhaps ask some interactive questions when the convention is unclear. Then that config could be committed so everyone would be on the same page.

pyflakes/checker.py
```diff
@@ -678,8 +678,30 @@ def GLOBAL(self, node):
...
+                self.messages = [
+                    m for m in self.messages if not
+                    isinstance(m, messages.UndefinedName) and not
+                    m.message_args[0] == node_name]
+
+                # Bind name to global scope if it doesn't exist already.
+                global_scope.setdefault(node_name, node_value)
+
+                # Bind name to non-global scopes, but as already "used".
+                node_value.used = True
```
asmeurer: @ryneeverett this line introduced a bug https://bugs.launchpad.net/pyflakes/+bug/1465824

ryneeverett: @asmeurer Fixed in #22.

PR #21, opened 2015-05-27
Add 'sudo: false' to .travis.yml

asmeurer:
This allows Travis to use the Docker container infrastructure, which should
make builds start up faster. See
http://docs.travis-ci.com/user/workers/container-based-infrastructure/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,3 +17,4 @@ script:
 matrix:
   allow_failures:
     - python: pypy
+sudo: false


asmeurer:
As you can see, the Travis CI build is already done. And I'm still waiting on builds from https://github.com/pyflakes/pyflakes/pull/20 to finish for commits that I pushed 15 minutes ago.

sigmavirus24:
Thanks @asmeurer. This is definitely a big improvement for contributors (so they can get faster results from Travis).

bitglue:
Nice change, thanks!

PR #22, opened 2015-06-16
Fix TypeError bug.

ryneeverett:
The Binding.used property must be a tuple (unless it's False).


bitglue:
Can we get a regression test?

ryneeverett:
@bitglue Done.

asmeurer:
I can confirm this fixes the bug on the file I was checking.

asmeurer:
For reference, this fixes https://bugs.launchpad.net/pyflakes/+bug/1465824.

bitglue:
Fixed in 0.9.2. Thanks!

PR #23, opened 2015-07-10
Fix a test under Python 3.5

myint:


sigmavirus24:
Looks good to me. I wish we could add 3.5 to travis but alas that takes some jumping through hoops.

bitglue:
I'm too lazy to install 3.5 to check, but this looks reasonable. Thanks for the patch.

PR #25, opened 2015-08-04
Different exit codes for different cases

binnisb:
To be able to run pyflakes as a script in a build pipeline for python code, it needs to differentiate between warnings and errors. Build pipelines should fail when there is a syntax error, but they can continue if there are only warnings from pyflakes.

Exit with code:
0 - No warnings or errors
1 - Only warnings
2 - Errors

Now one can do:
```bash
$ pyflakes module/
$ if [[ $? -eq 2 ]] 
...
```
but allow the code to pass if there are no errors, only warnings.

### DIFF
--- a/pyflakes/api.py
+++ b/pyflakes/api.py
@@ -172,4 +172,10 @@ def main(prog=None):
         warnings = checkRecursive(args, reporter)
     else:
         warnings = check(sys.stdin.read(), '<stdin>', reporter)
-    raise SystemExit(warnings > 0)
+    if reporter.errors:
+        code = 2
+    elif reporter.warnings:
+        code = 1
+    else:
+        code = 0
+    sys.exit(code)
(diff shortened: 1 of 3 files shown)


myint:
Just out of curiosity, why would the user want build system to find warnings, but then not report it with an exit status? And in terms of syntax error reporting, would `python -m compileall` fulfill your requirement?

By the way, I think the exit code of 2 is used by optparse (https://docs.python.org/3/library/optparse.html#parsing-arguments).

binnisb:
The build system should exactly do that. report warnings and errors, but
the warnings are not critical for the execution of the program, but having
syntax errors in your code is.

So, as a developer I want to get notified if my code will not run with exit
code X (I don't care what it is) and I want to get warnings of other types
with exit code Y.

I actually found a way to do this without introducing exit codes:
```bash
if [[ $({ pyflakes spotify 1> report } 2>&1 | wc -l ) -ne 0 ]] then;
echo "Codebreaking error"
else
echo "See repport for warnings"
fi
```

Cheers,
Brynjar

PR #26, opened 2015-08-14
Don't mark shadowing variable as 'used'.

ryneeverett:
Shadowing a used variable should still potentially raise an
UnusedVariable warning. Alter tests' expected results to agree with
this premise.

### DIFF
--- a/pyflakes/checker.py
+++ b/pyflakes/checker.py
@@ -96,10 +96,11 @@ class Binding(object):
                 line number that this binding was last used
     """
 
-    def __init__(self, name, source):
+    def __init__(self, name, source, isglobal=False):
         self.name = name
         self.source = source
         self.used = False
+        self.isglobal = isglobal
 
     def __str__(self):
         return self.name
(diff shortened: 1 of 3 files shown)


ryneeverett:
Travis is broken. This is incompatible with #27 so I'd happily rebase one if the other is merged.

bitglue:
What about a common idioms like this?

```console
$ pyflakes
try:
  from cStringIO import StringIO
except ImportError:
  import StringIO
<stdin>:4: 'StringIO' imported but unused
```

ryneeverett:
Ah, I figured those lines had to be for more than just `global`'s:

```py
if name in self.scope:
     binding.used = self.scope[name].used
```

It's also for when the name is being rebound in the context of a loop. As far as I can tell there's no reasonable way for pyflakes to detect this at present.

ryneeverett:
I think the best that can be done for now is to add tests and a comment. That will at least prevent me from opening an identical PR in a couple months.

bitglue:
Awesome, thanks for the PR.

PR #27, opened 2015-08-14
Don't report UnusedImport when binding global name

ryneeverett:
When an import binds to a name declared global by the "global"
statment, don't report it as unused.

### DIFF
--- a/pyflakes/checker.py
+++ b/pyflakes/checker.py
@@ -448,6 +448,10 @@ class Checker(object):
             elif isinstance(existing, Importation) and value.redefines(existing):
                 existing.redefined.append(node)
 
+        if value.name in self.scope:
+            # then assume the rebound name is used as a global or within a loop
+            value.used = self.scope[value.name].used
+
         self.scope[value.name] = value
 
     def getNodeHandler(self, node_class):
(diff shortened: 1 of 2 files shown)


ryneeverett:
Travis is broken. This is incompatible with #26 so I'd happily rebase one if the other is merged.

myint:
I'm not sure why, but Travis CI was toggled off. I've turned it on again. You can check the results now.

bitglue:
Can this please get rebased onto master? That should pick up the fix from #29 and give us success on Travis.

ryneeverett:
@bitglue Done.

bitglue:
Test looks good, and the goal seems sound, but I can't figure out how this works. Sometimes I just want to rewrite the whole thing :confounded: 

Anyone else?

sigmavirus24:
I had looked at it. I kind of wish there was a comment explaining that the ordering here was important, but I don't know it well enough to know why this works either.

ryneeverett:
I can't say I understand the reasoning behind these methods, but it seems like `handleNodeStore` is only called by the `NAME` method (which I'm assuming is only called for binding names with `foo = bar` syntax, whereas `addBinding` is called when a name is bound in another way too (imports, defining a function, defining a class, etc.).

ryneeverett:
> I kind of wish there was a comment explaining that the ordering here was important, but I don't know it well enough to know why this works either.

Rereading this comment I think I understand it today. The change doesn't have to do with the order. It is simply moving it into a more general method that is called for more statements, including `import`.

bitglue:
OK, let's go with it. Thanks for the patch.

PR #28, opened 2015-08-14
Use latest pep8

myint:
The one on PyPI does not support Python 3.5. The fix was made a month ago, but has not been released yet.

https://github.com/PyCQA/pep8/pull/420


myint:
I'm closing this as it introduces other problems. Maybe we should limit usage of `flake8` to one job.

PR #29, opened 2015-08-14
Avoid running flake8 under nightly

myint:
PyPI's pep8 is not compatible with Python 3.5 yet.


bitglue:
:+1: 

I'd prefer if you could squash the commits before merging.

myint:
Sure.

PR #30, opened 2015-08-26
Support Python 3.5 async/await statements for Pyflakes

playpauseandstop:
This PR add support of statements from [PEP 492](https://www.python.org/dev/peps/pep-0492/) to Pyflakes.

----

To test:

* Install Python 3.5.0rc2 as `pyenv install 3.5.0rc2`
* Set 3.5.0rc2 as default Python for the pyflakes as `pyenv local 3.5.0rc`
* Install tox, `pip install tox` and run tests as `tox -e py35`
* Check that everything works properly

----

I did not enabled `py35` in tox envlist as I think it's a little bit pre-mature.

### DIFF
--- a/pyflakes/checker.py
+++ b/pyflakes/checker.py
@@ -649,8 +649,9 @@ class Checker(object):
         pass
 
     # "stmt" type nodes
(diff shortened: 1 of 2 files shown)


sigmavirus24:
This looks good to me.

PR #31, opened 2015-08-31
Replace broken badge

myint:
`pypip.in` doesn't seem to work anymore. I'm not quite sure what benefit showing the wheel badge is, but here is a drop-in replacement.


bitglue:
I'm not sure what the benefit is, either. Want to just take it out?

myint:
Sure.

PR #37, opened 2015-11-02
Better del support

jayvdb:
Detect Python 3.2 and lower SyntaxError for deleting
variable referenced in nested scope.

### DIFF
--- a/pyflakes/checker.py
+++ b/pyflakes/checker.py
@@ -5,6 +5,7 @@ Implement the central Checker class.
 Also, it models the Bindings and Scopes.
 """
 import doctest
+import functools
 import os
 import sys
 
@@ -92,8 +93,8 @@ class Binding(object):
     which names have not. See L{Assignment} for a special type of binding that
     is checked with stricter rules.
 
-    @ivar used: pair of (L{Scope}, line-number) indicating the scope and
-                line number that this binding was last used
+    @ivar used: pair of (L{Scope}, node) indicating the scope and
+                the node that this binding was last used
     """
 
     def __init__(self, name, source):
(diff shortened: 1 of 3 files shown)


jayvdb:
Still needs a bit of polish, but I'd like confirmation that I am not wasting my time before proceeding with the current strategy.

A real world case is https://hg.python.org/cpython/file/tip/Lib/opcode.py

This would likely solve the original bug report in https://bugs.launchpad.net/pyflakes/+bug/1308508 .

(And https://bugs.launchpad.net/pyflakes/+bug/1461208 should be looked at as part of this)

The DoctestScope could be pulled out to be a separate changeset (or an alternative workaround found).  There are related bugs which could be solved as part of this; e.g. https://bugs.launchpad.net/pyflakes/+bug/1178807

I need to see whether the assignment checks can also be un-deferred, as that would reduce the amount of code restructure required to support del properly.

pyflakes/checker.py
```diff
@@ -326,11 +335,55 @@ def runDeferred(self, deferred):
...
-        for handler, scope, offset in deferred:
+        for record in deferred:
+            # node is an optional part of the tuple
+            # deferAssignment does not use it, and
+            # deferFunction does not require it
+            if len(record) == 3:
+                handler, scope, offset = record
+            else:
+                handler, scope, offset, node = record
+                if node:
```
sigmavirus24: `if node is not None`?

pyflakes/checker.py
```diff
@@ -92,8 +93,8 @@ class Binding(object):
     which names have not. See L{Assignment} for a special type of binding that
     is checked with stricter rules.
 
-    @ivar used: pair of (L{Scope}, line-number) indicating the scope and
-                line number that this binding was last used
+    @ivar used: pair of (L{Scope}, node) indicating the scope and
```
jayvdb: split off as https://github.com/pyflakes/pyflakes/pull/39

pyflakes/checker.py
```diff
@@ -240,6 +241,14 @@ class ModuleScope(Scope):
     pass
 
 
+# Doctest are similar to ModuleScope,
+# and on Python 3.2 and below need to allow
+# 'del' of nodes used in nested functions
+# like ModuleScope.
+class DoctestScope(FunctionScope):
```
jayvdb: split off as https://github.com/pyflakes/pyflakes/pull/38

bitglue:
Does this need a refresh after those other two PRs were split off?

jayvdb:
It certainly will need to be freshened, and a bit of polish.

PR #38, opened 2015-11-02
Add DoctestScope

jayvdb:


jayvdb:
This should fix https://bugs.launchpad.net/pyflakes/+bug/1512184

bitglue:
Is there a way we can add a test for this?

jayvdb:
I'd like ideas on how to effectively test this better.

`Checker` runs everything during it's constructor, which means an intrusive test harness is needed to make assertions regarding progress within the run.  A lot better tests could be constructed if `Checker.__init__` saved more of its state before returning.  e.g. putting a scope tree into a private variable.  Anyway, `Checker.deadScopes` provides a reasonable amount of detail to test the new `DoctestScope`.

More tests will be possible when the tuples in `self._deferredFunctions` contains `node` , which can be pulled out of https://github.com/pyflakes/pyflakes/pull/37 as a separate changeset.  `self._deferredFunctions` then contains pairs of scopes & node which can be verified are appropriate combinations.  `self._deferredFunctions` is also cleared before `__init__` returns, but that could be easily worked around with a test harness.

bitglue:
Well most of the tests are based on checking some snippet, then asserting that particular errors happened, or didn't. If an exception were to be raised in the checking of the snippet that would fail the test. Is there not a snippet that will elicit the bug being fixed here?

jayvdb:
The test `test_global_module_scope_pollution` verifies that the use of global is contained within the doctest scope and does not pollute the module scope.  That is what this section of code was trying to do, and that is what these new tests now confirm.

Anyway, I'll add some more tests which prevent this crash and similar crashes from slipping past code reviewers in the future.

bitglue:
It's looking pretty good. You seem to understand the mechanics of the problem better than I do :) The tests make it pretty clear though, which is good.

I guess all I'm looking for is a snippet that reproduces the traceback originally reported in the bug. The test could be as simple as calling `self.flakes()` on that code and then asserting that there were no errors, or there were errors, or whatever should happen. The important thing implicitly tested is that it didn't raise an exception.

bitglue:
Merged as 93aa3c4. Thanks!

jayvdb:
@bitglue, Why did you alter the commit?  I can not see any changes made in your amended commit.
And it was merge-able...?

bitglue:
I only altered the parent of the commit so there isn't a superfluous merge in the git history.

PR #40, opened 2015-11-08
PEP 498 f-strings support

jayvdb:
PEP 498 f-strings cause pyflakes to crash with
AttributeError: 'Checker' object has no attribute 'FORMATTEDVALUE'

### DIFF
--- a/pyflakes/test/test_other.py
+++ b/pyflakes/test/test_other.py
@@ -989,6 +989,14 @@ class TestUnusedAssignment(TestCase):
         """Do not crash on lone "return"."""
         self.flakes('return 2')
 
+    @skipIf(version_info < (3, 6), 'new in Python 3.6')
+    def test_f_string(self):
+        """Test PEP 498 f-strings are treated as a usage."""
+        self.flakes('''
+        baz = 0
+        print(f'\x7b4*baz\N{RIGHT CURLY BRACKET}')
+        ''')
+
 
 class TestAsyncStatements(TestCase):
 
--- a/pyflakes/checker.py
+++ b/pyflakes/checker.py
@@ -675,7 +675,7 @@ class Checker(object):
         EQ = NOTEQ = LT = LTE = GT = GTE = IS = ISNOT = IN = NOTIN = ignore
 
     # additional node types
-    COMPREHENSION = KEYWORD = handleChildren
+    COMPREHENSION = KEYWORD = FORMATTEDVALUE = handleChildren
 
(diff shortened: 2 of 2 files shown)


asmeurer:
Hey, I didn't see that that PEP was accepted. Really, it should mark variables as used if they are in an f-string. But Python 3.6 won't be released for some time, so there's no rush.

jayvdb:
That PEP is implemented in Python 3.6.  Agreed it isnt critical to provide Python 3.6 support yet, as it is still alpha & unreleased.
This changeset does cause the variables in the f-string to be marked as used ( FORMATTEDVALUE is an expression, with children ): see the test case, which would fail if the children were not processed.

asmeurer:
Oh I didn't notice that. Why do you use escape sequences instead of just `{}`? That was throwing me off.

jayvdb:
Well it is the most complex case.  Using escape sequences ensures that the test will break if anyone in the future decides that pyflakes should to parse the f-string to determine the names used.

It is the same syntax shown in
https://www.python.org/dev/peps/pep-0498/#escape-sequences

Happy to make it simple if requested.

bitglue:
I don't use Python 3.6, but if this avoids a crash it seems an improvement to merge it as is. Thanks for the patch.

pyflakes/test/test_other.py
```diff
@@ -989,6 +989,14 @@ def test_returnOnly(self):
         """Do not crash on lone "return"."""
         self.flakes('return 2')
 
+    @skipIf(version_info < (3, 6), 'new in Python 3.6')
+    def test_f_string(self):
+        """Test PEP 498 f-strings are treated as a usage."""
+        self.flakes('''
+        baz = 0
+        print(f'\x7b4*baz\N{RIGHT CURLY BRACKET}')
```
asottile: why the strange syntax here?

The strange syntax doesn't actually affect the test since it gets evaluated **before** pyflakes evaluates it:

```pycon
>>> '''
...         baz = 0
...         print(f'\x7b4*baz\N{RIGHT CURLY BRACKET}')
... '''
"\n        baz = 0\n        print(f'{4*baz}')\n"
```

And if you wanted to _literaly_ check that, it wouldn't be a valid f-string placeholder:

```pycon
>>> baz = 0
>>> print(f'\x7b4*baz\N{RIGHT CURLY BRACKET}')
{4*baz}
```

This should just be written as `{4*baz)` I imagine?

PR #41, opened 2015-11-12
Fix global removing all UndefinedName

jayvdb:
03ffc763 introduced better global support,
however a logic error caused it to remove any
previously detected UndefinedName whenever
a global was encountered.


bitglue:
good catch, thanks for the patch

ryneeverett:
Very nice.

PR #44, opened 2015-11-15
Import in Class is a public member

jayvdb:
An import in a class is a member of the class,
and may be used outside the class scope, so it
can not be marked as unused.

Resolves lp:1463566


spaceone:
Thank you very much!!

PR #45, opened 2015-11-15
Always report imports shadowed by another import

jayvdb:
Previously only unused imports redefined by another
imports were reported as RedefinedWhileUnused.

A new ImportShadowedByImport message now always
reports the re-import except when an import in the
doctest scope shadows an import in the module scope.

### DIFF
--- a/pyflakes/test/test_imports.py
+++ b/pyflakes/test/test_imports.py
@@ -184,9 +184,8 @@ class Test(TestCase):
             def baz():
                 def fu():
                     pass
-        ''',
-                    m.RedefinedWhileUnused, m.RedefinedWhileUnused,
-                    m.UnusedImport, m.UnusedImport)
+        ''', *(m.ImportShadowedByImport, m.RedefinedWhileUnused,
+               m.UnusedImport, m.UnusedImport))
 
     def test_redefinedButUsedLater(self):
         """
@@ -563,6 +562,15 @@ class Test(TestCase):
             import fu
         fu
         ''', m.UnusedImport, m.UndefinedName)
+
+    def test_shadowedInNestedScope(self):
+        self.flakes('''
+        import fu
+        def bar():
+            import fu
+            fu
+        fu
+        ''', m.ImportShadowedByImport)
 
     def test_methodsDontUseClassScope(self):
         self.flakes('''
--- a/pyflakes/checker.py
+++ b/pyflakes/checker.py
@@ -449,8 +449,14 @@ class Checker(object):
                     self.report(messages.RedefinedWhileUnused,
                                 node, value.name, existing.source)
 
-            elif isinstance(existing, Importation) and value.redefines(existing):
-                existing.redefined.append(node)
+            elif isinstance(existing, Importation):
+                if isinstance(node, (ast.Import, ast.ImportFrom)):
+                    if not isinstance(self.scope, DoctestScope):
+                        self.report(messages.ImportShadowedByImport,
+                                    node, value.name, existing.source)
+
+                elif value.redefines(existing):
+                    existing.redefined.append(node)
 
         if value.name in self.scope:
             # then assume the rebound name is used as a global or within a loop
--- a/pyflakes/messages.py
+++ b/pyflakes/messages.py
@@ -41,6 +41,14 @@ class RedefinedInListComp(Message):
         self.message_args = (name, orig_loc.lineno)
 
 
+class ImportShadowedByImport(Message):
+    message = 'import %r from line %r shadowed by import'
+
+    def __init__(self, filename, loc, name, orig_loc):
+        Message.__init__(self, filename, loc)
+        self.message_args = (name, orig_loc.lineno)
+
+
 class ImportShadowedByLoopVar(Message):
     message = 'import %r from line %r shadowed by loop variable'
 
--- a/pyflakes/test/test_doctests.py
+++ b/pyflakes/test/test_doctests.py
@@ -9,7 +9,7 @@ from pyflakes.checker import (
 from pyflakes.test.test_other import Test as TestOther
 from pyflakes.test.test_imports import Test as TestImports
 from pyflakes.test.test_undefined_names import Test as TestUndefinedNames
-from pyflakes.test.harness import TestCase, skip
+from pyflakes.test.harness import TestCase
 
 
 class _DoctestMixin(object):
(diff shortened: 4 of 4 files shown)


pyflakes/test/test_imports.py
```diff
@@ -184,9 +184,8 @@ def bar():
             def baz():
                 def fu():
                     pass
-        ''',
-                    m.RedefinedWhileUnused, m.RedefinedWhileUnused,
-                    m.UnusedImport, m.UnusedImport)
+        ''', *(m.ImportShadowedByImport, m.RedefinedWhileUnused,
```
bitglue: What's the use of the `*(...)` construction here?

pyflakes/test/test_imports.py
```diff
@@ -564,6 +563,15 @@ def bar():
...
         ''', m.UnusedImport, m.UndefinedName)
 
+    def test_shadowedInNestedScope(self):
+        self.flakes('''
+        import fu
+        def bar():
+            import fu
+            fu
+        fu
+        ''', m.ImportShadowedByImport)
```
bitglue: I'm not sure I agree this is a desirable behavior. I know there's also an `ImportShadowedByLoopVar` ... I'm not sure I agree with that either. Citing the [design principles](https://github.com/pyflakes/pyflakes#design-principles):

> Pyflakes makes a simple promise: it will never complain about style, and it will try very, very hard to *never emit false positives*.

There's nothing inherently erroneous about the code here. Can we be sure that no one will ever have a legitimate reason to do this? What if we change the example a bit:

```python
from StringIO import StringIO

def bar():
    from cStringIO import StringIO
    StringIO

StringIO
```

jayvdb: Unused imports is also style, not an error.

I could exlude ast.ImportFrom from this new rule to avoid some scenarios where redefining an imported name is less problematic, such as your example.

pyflakes/test/test_imports.py
jayvdb: To simplify neater indentation.

pyflakes/test/test_imports.py
bitglue: That could work. Though, #44 makes me think of another case:

```python
import foo

class bar:
    import foo

foo
```

Probably there should be no errors here, but right now it looks like the 2nd `import foo` yields "test.py:4: import 'foo' from line 1 shadowed by import"

jayvdb:
Yea, I was going to note that this and #44 should not both be merged without amendments, but since you require fast-forward merges only that note isnt necessary.

pyflakes/test/test_imports.py
jayvdb: Differentiating between types of imports will be much cleaner and simpler after #54

bitglue:
Sorry I haven't been keeping up with the PRs lately. I'm busy with work stuff this week but I'll try to take a look at these next week. Alternately, I'm fine with any of these changes being merged if anyone else reviews them first.

pyflakes/test/test_imports.py
asmeurer: But this isn't even bad style. It's no different from

```
var = 1

def func():
    var = 2
    ...
```

Redefining variables in inner scopes is a perfectly legitimate thing to do, whether that be via imports, loops, `as` constructs, or assignments. Python's scoping rules mostly protect you from this doing anything bad.

jayvdb: IMO it isnt the same, as imports are an external resource, and should be consistently named through the module.  I can appreciate if others think that is too much a style issue for pyflakes to take on.

However, the case I am most interested in is

```
import foo

def blah():
     import foo
     foo

foo
```

Currently that isnt caught as a redefinition because `foo` is used.  However the import in the function is superfluous, and should be removed.

pyflakes/test/test_imports.py
sigmavirus24: Eh, this isn't exactly clear or obvious. I understand why, but I'm not sure I agree with it.

jayvdb: No worries; I can rework it when I create a narrower patch after #54 has been approved.

ambv:
Similarly to #55, I would use this but it doesn't belong to pyflakes on the basis of not reporting an actual error. A superfluous import is absolutely harmless and a local, alternative import is useful (even if constitutes questionable style).

I do agree unused imports break the "errors only" rule, too, but this is historical and it seems to me it would open a Pandora box of other warning-type suggestions if I let this one in. It really seems we need a flake8-likely-bugs plugin that sits between what pyflakes tests for (hard errors) and what pep8 does (pure style suggestions).

bitglue:
Superfluous imports aren't harmless: they have (sometimes significant) performance consequences and create edges in the dependency graph that are unnecessary which can lead to dependency cycles that cause imports to fail in confusing and mysterious ways.

In fact this very problem was the reason pyflakes was written in the first place.

I think you miss the point of pyflakes. It is not to catch "hard errors". That's what unit tests are for. It's to catch "soft errors" which would not be caught by a unit test, but only *actual errors*, not *maybe errors*.

There is never any reason to import something and not use it, with one exception: importing stuff into __init__.py or similar to make an ubermodule that consolidates the namespaces of multiple submodules. And this is why we have `__all__`.

asmeurer:
There are use-cases (like when a module import doesn't exist until you import some other thing, like https://twitter.com/asmeurer/status/468514695154921472), but they are very rare, and I think generally can be worked around. I agree the unused imports, while they seem harmless, are actually useful, in the same way that unused variables are useful. If you imported something and don't use it, that often indicates that you forgot something (not as often as importing a variable and not using it, but often enough). I've also had pyflakes tell me that I'm not using an external module any more, meaning my code had an unnecessary unused dependency from an unused import. And unused imports are trivial to fix, and I doubt anyone will argue against doing so for some style reason (contrast this with some pep8 warnings, some of which I strongly disagree with). And FWIW I disagree that you should have to define `__all__` in `__init__.py` for pyflakes to not warn about unused imports/names.

ambv:
@bitglue, by a superfluous import I mean the exact case covered by this pull request, e.g. an import done within the function scope, when a module-level import has already been made. These have almost zero runtime penalty as they're reusing modules already present in `sys.modules`.

pyflakes/test/test_imports.py
jayvdb: Just noting that in the case of the class attribute, the same can be achieved without shadowing using

```
import foo

class bar:
    foo = foo

```

But as said earlier, I am happy to allow shadowing for class attributes, and other instances if there is consensus one way or the other.

PR #48, opened 2015-11-18
Python 3 only allows import * at module level

jayvdb:


bitglue:
What's the use case here? I'd think since any use of import *, anywhere, in any version, already is reported by PyFlakes, the previous behavior is fine. Why do we need to change the error that it is?

asmeurer:
It's a syntax error, not just a warning. The current warning can be ignored if you know what you're doing, but a syntax error cannot be ignored.

PR #49, opened 2015-11-19
Fix undefined name in generators in class

jayvdb:
8c8a27b8 provided a partial solution for generators in the class scope,
however it did not account for generators enclosing other generators.


bitglue:
Looks good. Thanks for all the contributions.

Would you please rebase? If you also want to include the commit from #50 then I can fast-forward merge both in one go.

PR #50, opened 2015-11-19
Fix PyPy

jayvdb:
Resolves lp:1186139


bitglue:
Looks good. Rebase please?

PR #51, opened 2015-11-20
Report assert using tuple

jayvdb:
This is a SyntaxWarning on Python 2.6+

Resolves lp:848467

### DIFF
--- a/pyflakes/checker.py
+++ b/pyflakes/checker.py
@@ -671,7 +671,7 @@ class Checker(object):
 
     # "stmt" type nodes
     DELETE = PRINT = FOR = ASYNCFOR = WHILE = IF = WITH = WITHITEM = \
-        ASYNCWITH = ASYNCWITHITEM = RAISE = TRYFINALLY = ASSERT = EXEC = \
+        ASYNCWITH = ASYNCWITHITEM = RAISE = TRYFINALLY = EXEC = \
         EXPR = ASSIGN = handleChildren
 
     PASS = ignore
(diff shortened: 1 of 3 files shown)


pyflakes/checker.py
```diff
@@ -697,6 +697,11 @@ def ignore(self, node):
     # additional node types
     COMPREHENSION = KEYWORD = FORMATTEDVALUE = handleChildren
 
+    def ASSERT(self, node):
+        if isinstance(node.test, ast.Tuple):
```
bitglue: What about

```python
assert ()
```

Not sure why anyone would do that instead of just `assert False`, but it's always False (so has some utility), and doesn't elicit a SyntaxWarning on Python 2.7, so we probably shouldn't complain about it.

jayvdb: Nice catch.  I've excluded that bizarre case.

You may have seen that I also posted on https://bugs.launchpad.net/pyflakes/+bug/848467 , linking to an expanded solution which trials detecting static asserts (which is where `assert ()`, should be reported, if desirable)

pyflakes/test/test_other.py
```diff
@@ -1633,6 +1633,40 @@ def test_augmentedAssignmentImportedFunctionCall(self):
...
+
+    def test_assert_tuple(self):
+        """An assert of a tuple is always True."""
+        self.flakes('''
+        assert (False, 'x')
+        assert (False, )
+        ''', m.AssertTuple, m.AssertTuple)
+
+    def test_assert_tuple_empty(self):
+        """An assert () is always True but should not report AssertTuple."""
```
bitglue: `()` is always False, right?

Otherwise, :+1:

PR #52, opened 2015-11-22
Process doctest scope directly under the module scope

jayvdb:
Explicitly place the doctest directly under the module scope,
instead of processing it while within the scope where the
docstring was encountered.

The primary benefit of moving the doctest scope directly under
the module scope is that it may efficiently be found, as it
may only appear second in the stack.

However it also ensures that the doctest scope can not access
names in scopes between the module scope and the object where the
docstring was encountered.  As it was, class scope rules prevented
the doctest scope from accessing names in a class scope, however
the doctest scope was able to access names in an outer function,
if the doctest appeared in a nested function.
Note that there was no real bug there, as the doctest module does
not process doctest in nested functions (Python issue #1650090),
so doctest appearing in nested functions are informational only.
pyflakes does inspect doctest in nested functions, and now it
ensures they can not access names in an outer function.

### DIFF
--- a/pyflakes/test/test_doctests.py
+++ b/pyflakes/test/test_doctests.py
@@ -190,6 +190,54 @@ class Test(TestCase):
             '''
         """, m.UndefinedName)
 
+    def test_nested_class(self):
+        """Doctest within nested class are processed."""
+        self.flakes("""
+        class C:
+            class D:
+                '''
+                    >>> m
+                '''
+                def doctest_stuff(self):
+                    '''
+                        >>> m
+                    '''
+                    return 1
+        """, m.UndefinedName, m.UndefinedName)
+
+    def test_ignore_nested_function(self):
+        """Doctest module does not process doctest in nested functions."""
+        # 'syntax error' would cause a SyntaxError if the doctest was processed.
+        # However doctest does not find doctest in nested functions
+        # (https://bugs.python.org/issue1650090). If nested functions were
+        # processed, this use of m should cause UndefinedName, and the
+        # name inner_function should probably exist in the doctest scope.
+        self.flakes("""
+        def doctest_stuff():
+            def inner_function():
+                '''
+                    >>> syntax error
+                    >>> inner_function()
+                    1
+                    >>> m
+                '''
+                return 1
+            m = inner_function()
+            return m
+        """)
+
+    def test_inaccessible_scope_class(self):
+        """Doctest may not access class scope."""
+        self.flakes("""
+        class C:
+            def doctest_stuff(self):
+                '''
+                    >>> m
+                '''
+                return 1
+            m = 1
+        """, m.UndefinedName)
+
     def test_importBeforeDoctest(self):
         self.flakes("""
         import foo
--- a/pyflakes/checker.py
+++ b/pyflakes/checker.py
@@ -254,11 +254,11 @@ class GeneratorScope(Scope):
 
 
 class ModuleScope(Scope):
-    pass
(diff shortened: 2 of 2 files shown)


ryneeverett:
Shouldn't the `inner_function` be accessible in its own doctest? It no longer is now that the doctest is in global scope. That is, the following would now raise `UndefinedName`:

```py
        self.flakes("""
        def doctest_stuff():
            def inner_function():
                '''
                    >>> inner_function()
                '''
                return 1
            m = inner_function()
            return m
        """)
```

jayvdb:
I agree it would be useful if `inner_function` was included in the doctest scope, and that should discussed on https://bugs.python.org/issue1650090 .  It is easy on the pyflakes side, to add the name into the DoctestScope.

Maybe we shouldnt process the doctest in nested functions, as you've pointed out there is a chance that this changeset would cause new errors.  And those new errors couldnt easily be resolved, as we dont know yet how doctest in nested functions will be implemented.

ryneeverett:
> Maybe we shouldnt process the doctest in nested functions

Makes sense to me. There might even be a case for adding an error message for nested doctests since they wont run.

jayvdb:
>> Maybe we shouldnt process the doctest in nested functions

> Makes sense to me.

I've implemented that, and added tests.

> There might even be a case for adding an error message for nested doctests since they wont run.

That is easy to add, if @bitglue agrees it is OK.

pyflakes/test/test_doctests.py
```diff
@@ -190,6 +190,50 @@ def doctest_stuff():
...
+            def inner_function():
+                '''
+                    >>> inner_function()
+                    1
+                    >>> m
+                '''
+                return 1
+            m = inner_function()
+            return m
+        """)
```
ryneeverett: Would a simpler test suffice? For example a would-be SyntaxError:

        self.flakes('''
        def doctest_stuff():
            def inner_function():
                """
                    >>> if True:
                    ... pass
                """
        ''')

ryneeverett:
> > There might even be a case for adding an error message for nested doctests since they wont run.

> That is easy to add, if @bitglue agrees it is OK.

I'm having second thoughts. I guess I was initially thinking we'd have to skip them with an `if: pass` and if we're doing that we might as well raise an error. But you did it a better way and I'm not sure adding the code is worth the maintenance.

pyflakes/test/test_doctests.py
jayvdb: A simpler test would be good.  I'd go with a slightly more obvious syntax error, which doesnt use doctest indentation, like `syntax error` or `if: pass`.  Is there a more brutally obvious syntax error?

I'd prefer to keep the existing use of 'inner_function' and 'm', so that if/when nested functions are processed by doctest, those variables will be re-considered, if only because git blame will point the coder back to this discussion.  But I'm happy to remove them.

ryneeverett: > A simpler test would be good. I'd go with a slightly more obvious syntax error, which doesnt use doctest indentation, like syntax error or if: pass. Is there a more brutally obvious syntax error?

I like `syntax error` myself.

> I'd prefer to keep the existing use of 'inner_function' and 'm', so that if/when nested functions are processed by doctest, those variables will be re-considered, if only because git blame will point the coder back to this discussion. But I'm happy to remove them.

I don't really care as long as the error is more obvious.

jayvdb: I've added 'syntax error', and a more detailed comment.

PR #53, opened 2015-11-24
Report unused assignments in doctest

jayvdb:
Assignment to '_' is not reported as unused.


pyflakes/checker.py
```diff
@@ -254,11 +262,15 @@ def __init__(self):
...
+
+        All assignments are considered used when a scope uses locals().
+
+        Unused names in alwaysUsed are skipped.
         """
-        for name, binding in self.items():
-            if (not binding.used and name not in self.globals
-                    and not self.usesLocals
-                    and isinstance(binding, Assignment)):
+        if self.usesLocals:
```
jayvdb: This should wait until https://github.com/PyCQA/pyflakes/pull/343

PR #54, opened 2015-11-26
Importation classes with imported name and alias

jayvdb:
In order to solve many corner cases related to imports,
more information is needed about each import.

This change creates two new classes:
- SubmoduleImportation
- ImportationFrom

And adds an optional parameter full_name to the super class
Importation.

Functionally, this change only improves existing error messages
to report the full imported name where previously an error
would include only the import alias.

### DIFF
--- a/pyflakes/checker.py
+++ b/pyflakes/checker.py
@@ -136,16 +136,83 @@ class Importation(Definition):
     @type fullName: C{str}
     """
 
-    def __init__(self, name, source):
-        self.fullName = name
+    def __init__(self, name, source, full_name=None):
+        self.fullName = full_name or name
         self.redefined = []
-        name = name.split('.')[0]
         super(Importation, self).__init__(name, source)
+
+    def redefines(self, other):
+        return isinstance(other, Definition) and self.name == other.name
+
+    def _has_alias(self):
+        """Return whether importation needs an as clause."""
+        return not self.fullName.split('.')[-1] == self.name
+
+    @property
+    def source_statement(self):
+        """Generate a source statement equivalent to the import."""
+        if self._has_alias():
+            return 'import %s as %s' % (self.fullName, self.name)
+        else:
+            return 'import %s' % self.fullName
+
+    def __str__(self):
+        """Return import full name with alias."""
+        if self._has_alias():
+            return self.fullName + ' as ' + self.name
+        else:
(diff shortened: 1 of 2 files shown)


jayvdb:
`source_statement` exists as I would like to include that in a better `__repr__`

Note that I have a functionally equivalent changeset which does all the work within the existing `Importation` class: https://github.com/jayvdb/pyflakes/commit/e250ea6085a15485889115b8524d616254e127aa . IMO that isnt as neat and tidy; the special case for `_submodule_import` adds complexity throughout.

jayvdb:
@bitglue , could you please review this.  I have pending changes that depend on this.

bitglue:
Looks good on a brief read. I'll take a deeper look tonight.

bitglue:
This looks good. Can you please rebase onto master?

pyflakes/checker.py
```diff
@@ -136,16 +136,83 @@ class Importation(Definition):
...
+            return self.fullName
+
+
+class SubmoduleImportation(Importation):
+
+    def __init__(self, name, source):
+        # A dot should only appear in the name when it is a submodule import
+        # without an 'as' clause, which is a special type of import where the
+        # root module is implicitly imported, and the submodules are also
+        # imported by Python does not restrict which attributes of the root
```
jayvdb: ah, just noticed : 'by' should be 'but' or 'however'

pyflakes/checker.py
```diff
@@ -136,16 +136,83 @@ class Importation(Definition):
...
+        """Generate a source statement equivalent to the import."""
+        if self._has_alias():
+            return 'import %s as %s' % (self.fullName, self.name)
+        else:
+            return 'import %s' % self.fullName
+
+    def __str__(self):
+        """Return import full name with alias."""
+        if self._has_alias():
+            return self.fullName + ' as ' + self.name
```
jayvdb: would we prefer `%s` syntax here instead of `+` ?  It does seem that %s syntax is preferred.

bitglue: I don't have any particular preference.

collinanderson:
This appears to break relative imports like: `from . import urls`

jayvdb:
Yea, there were no tests covering them, and I forgot about those.  The fix #61 is waiting for review.

PR #60, opened 2016-04-24
RFC: checking call sites for invalid keyword parameters

alon:
Hi,

 This patchset is RFC quality, I'm submitting it not for inclusion but for the discussion. What I want to support is checking of call sites between modules. The patch only does the check for the same module, i.e. the added test shows the usage:

```python
def f(x):
   pass

f(y=10) # will report a wrong call, y is invalid, only x is possible
```

But I really want to be able to support the same test with f defined in foo:

```python
# g.py
from foo import f
f(y=10) # report same error

# foo.py
def f(x)
  pass
```

To support that I'd have to break pyflakes one file at a time parsing. I can do it completely optionally, i.e. only when called on a group of files, so it would not break and won't add any runtime requirements on existing users.

Would that be something that anyone would be interested in?

### DIFF
--- a/pyflakes/checker.py
+++ b/pyflakes/checker.py
@@ -726,7 +726,6 @@ class Checker(object):
     def handleCall(self, call):
         # checks only keywords for now.
         # TODO - check positional arguments as well
-        used_keywords = {k.arg for k in call.keywords}
         function_name = getNodeName(call.func)
         assert function_name is not None # 'attr' in call.func._fields ?
         try:
@@ -735,9 +734,22 @@ class Checker(object):
             self.report('{} not in scope {}, error'.format(function_name, self.scope))
             return
         fdef = func_binding.source
+        if isinstance(fdef, ast.FunctionDef):
+            self.validateCall(fdef, call)
+        elif isinstance(fdef, ast.Import):
+            self.validateImportCall(fdef, call)
+        else:
+            assert False, "unexpected call binding: not Import, not FunctionDef"
+
+    def validateImportCall(self, importdef, call):
+        print("TODO")
+
+    def validateCall(self, fdef, call):
         if fdef.args.kwarg is not None:
             # ignore calls to double star functions
             return
+        used_keywords = {k.arg for k in call.keywords}
+        function_name = getNodeName(call.func)
         valid_keywords = {arg.id for arg in fdef.args.args}
         for k in used_keywords - valid_keywords:
             self.report(messages.InvalidKeywordInCall, call, function_name=function_name,
--- a/pyflakes/checker.py
+++ b/pyflakes/checker.py
@@ -362,6 +362,14 @@ def getNodeName(node):
         return node.name
 
 
+def getCallName(call):
+    node = call.func
+    if hasattr(node, 'id'):
+        return node.id
+    if hasattr(node, 'attr'):
+        return node.attr
+
+
 class Checker(object):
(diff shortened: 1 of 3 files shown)


bitglue:
> To support that I'd have to break pyflakes one file at a time parsing. I can do it completely optionally, i.e. only when called on a group of files, so it would not break and won't add any runtime requirements on existing users.

Provided performance is not significantly decreased, and code is only parsed but not executed, I think this would be really great.

jayvdb:
Loading modules dependencies is an obvious step for pyflakes to take. The first step would be validating imported symbols exist in the module being imported. That needs to be merged and stable before validating calls between modules.

Regarding this patch as it is without intermodule support, I'd like to know if/how you intend to handle calls of methods, i.e.`x.y()` .

I'm not sure this is suitable to be activated by default if it only handles functions and not class attributes being called, but we could add that afterwards if it isnt an itch you want to scratch.

asmeurer:
Method calls are a can of worms. So much metaprogramming can and does happen with them (not to mention mro dispatch).

bitglue:
> Method calls are a can of worms. So much metaprogramming can and does happen with them (not to mention mro dispatch).

Probably, but I'd guess there are at least some cases where some reasonable checking can be done. There's always some level of trickery that will confuse pyflakes, like manipulating locals(). But those things are usually contrived, so in practice they aren't a problem.

I think the trick will be detecting when common things (decorators, metaclasses) are being used, and avoiding false positives in those cases.

jayvdb:
How would we handle (function) calls to stdlib, and external libs? They would be out of scope according to the initial RFC description.  I'd like to see stdlib call checking in scope, and suspect external libs should be out of scope as that requires a lot of path loading logic and complexity.

alon:
Since I submitted this RFC I discovered mypy does what I want already, so I'm less inclined to finish this work. Anyway I appreciate the discussion, and maybe I will complete it after all, since pyflakes is much simpler and runs with python 2.7 so is easier to use in my project (mypy has a python 2 mode but requires python 3 to run since it is not compatible with python 2).

asmeurer:
For builtins and stdlib you could just import them and check them. Importing a lot of the standard library could slow down pyflakes, but it shouldn't result in any side effects as far as I know.

jayvdb:
The best place to start with detecting names from module imports is star imports, which currently cause lots of messages like "x may be undefined, or defined from star imports: y" .

The performance penalty of handling the star imported module is certainly appropriate as the author has chosen to use star imports, and the cost of handling the star imported module has the direct benefit: the existing error message can be silenced if the name exists in the star imported module, or a better error message can be presented if the symbol exists in multiple star imported modules.

asmeurer:
Well * imports are already bad practice, so I don't mind so much pyflakes bugging about them. But I would actually *love* if it could support them in `__ini__.py` files. Right now pyflakes forces a particular style in `__init__.py` that almost no one uses. 

For performance, if someone already runs `pyflakes module/`, then there's no performance lost if pyflakes checks the whole module (assuming it does it efficiently, i.e., without re-reading files). Ultimately it may need to be controllable via a command line flag (or environment variable or <insert your favorite method of configuration here>).

bitglue:
Since the author of this PR has no intention to complete it, I'm closing it. Feel free to continue the discussion though, I think the concept is sound if someone wants to try implementing it.

PR #62, opened 2016-05-04
Suppress RedefinedWhileUnused for submodule import

jayvdb:
Fixes lp:1578051

aec68a7 added module names to error messages, which included
a new class SubmoduleImportation to handle the special case
of submodule imports.  It correctly handled the case of
a submodule import occurring after the root module was imported,
but didnt handle the opposite case of the submodule import
occurring before the root module was imported.

### DIFF
--- a/pyflakes/checker.py
+++ b/pyflakes/checker.py
@@ -142,6 +142,9 @@ class Importation(Definition):
         super(Importation, self).__init__(name, source)
 
     def redefines(self, other):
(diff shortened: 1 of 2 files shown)


jayvdb:
@sigmavirus24 , this is also a regression in the last release.  A review would be helpful, so I can merge it.

sigmavirus24:
Thanks @jayvdb

PR #67, opened 2016-05-06
Avoid traceback when exception is del-ed in except

bitglue:
Fixes a regression introduced by
2a698f87c02a43d4489e30481e9def14ed4b4431.

This would fail with a KeyError:

try:
    pass
except Exception as e:
    del e

Fixes lp:1578903

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 *.egg-info
 *.pyc
 .tox
+/.cache/
 /dist/
 /build/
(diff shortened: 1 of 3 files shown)


jayvdb:
:+1:

I've been testing with an if block instead of a try except, as I cant see how we might benefit from a try except.

but both fix the bug.

sigmavirus24:
:+1:

PR #68, opened 2016-05-06
Manage recursion limit preventing RuntimeError

jayvdb:
pyflakes has traditionally recursed with a handler for every
level of the ast.  The ast depth can become very large, especially
for an expression containing many binary operators.

Python has a maximum recursion limit, defaulting to a low number
like 1000, which resulted in a RuntimeError for the ast of:

    x = 1 + 2 + 3 + ... + 1001

To workaround this problem, pyflakes now increases the recursion
limit at runtime when it knows it will be exceeded.

Fixes lp:1507827


jayvdb:
I've gone for the simple workaround for https://bugs.launchpad.net/pyflakes/+bug/1507827 , as this is the only Exception I can find, and the alternative (flattening) is a complex patch not suitable for a patch or even minor release.

I'd really like this solved, as it means I can run `pyflakes .` in a parent directory above all of my python packages as a form of regression testing, but that fails currently due to this exception.

I have one more minor bug in my flatten patch, but should have it up in a few hours.  It is easy enough to verify for unary/boolean/binary ops only, but the bug would still exist elsewhere.  Solving it for boolean/binary op is not easy, as boolean/binary ops can contain lambda's, and pyflakes has deferred processing of lambdas.  I've fixed that problem - my issue is now just generators on py2.7 only.  What I do know is the fix is going to need a lot of vetting & testing before it is released.

jayvdb:
#70 is the companion patch, which reduces the recursion so it doesnt hit the limit easily.  It still can hit the limit for extremely crazy code, but it is quite improbable.

I'm going to slightly refactor this patch (68) so it is less of a performance hit, so it can be merged into the 1.2. or 1.x series promptly.

pyflakes/test/test_other.py
```diff
@@ -1993,3 +1994,18 @@ def test_raise_notimplemented(self):
...
+
+
+class TestMaximumRecursion(TestCase):
+
+    def setUp(self):
+        self._recursionlimit = sys.getrecursionlimit()
+
+    def test_recursion_limit(self):
+        # Using self._recursionlimit * 10 tends to cause CPython to core dump.
+        r = range(self._recursionlimit * 9)
```
jayvdb: Lots of segmentation fault now after a rebase, on `self._recursionlimit * 9`

I'll try reducing it.

jayvdb:
The test is failing on older versions of PyPy only.
Leaving this open as it does work, and is slightly faster than https://github.com/PyCQA/pyflakes/pull/70 .
If it is preferred, we could skip the tests on those environments where pushing the runtime beyond the limit causes core dumps.

PR #69, opened 2016-05-07
Run pyflakes against all code in the virtualenv

jayvdb:


jayvdb:
Here is what it looks like after the relative-import fix is merged: https://travis-ci.org/jayvdb/pyflakes/builds/128447040

jayvdb:
Note that complete log is nicely 'folded'/hidden [on Python 3.5+](https://travis-ci.org/jayvdb/pyflakes/jobs/128447046) as it has less flakes and thus fits within the allowed size for the log in the Travis web UI.

jayvdb:
just rebased.

sigmavirus24:
Can you explain *why* this is necessary? I feel like this is going to add a lot of noise in the travis logs.

bitglue:
The answer to the "why" question should probably go in the commit log message for posterity

jayvdb:
It would have blocked both of the regressions that occurred in 1.2 .

sigmavirus24:
@bitglue is correct. The *why* for this should be in the commit log and the file should have a comment telling people this is meant to prevent regressions that will only be caught on Travis (not locally). That comment will ideally cause people to look at the commit log and see your reasoning for this.

asottile:
I'm planning to make an actual framework for tracking differences which would obsolete this

PR #70, opened 2016-05-07
Flatten handlers

jayvdb:
pyflakes has traditionally recursed with a handler for every
level of the ast.  The ast depth can become very large, especially
for an expression containing many binary operators.

Python has a maximum recursion limit, defaulting to a low number
like 1000, which resulted in a RuntimeError for the ast of:

    x = 1 + 2 + 3 + ... + 1001

This change avoids recursing for nodes that do not have a specific
handler.

Checker.nodeDepth and node.depth change from always being the
ast depth, which varied based on Python version due to ast differences,
to being the number of nested handlers within pyflakes.


jayvdb:
This is just a WIP; while passes all tests, it is very likely that more tests are needed, and some of the handlers need to be switched back to being `handleChildren` instead of `handleChildrenFlattened`.

jayvdb:
Not really a WIP any more.  tests now exist for every case which didnt fail when they should have before.  I've run it against against ~300 of the largest OSS Python codebases, and it has no differences except ordering of errors within the same line.

A slightly older version of this change had a minor perf hit, of 3-5 mins increase for what was a 55 min task of processing all of those ~300 OSS codebases .  I expect that perf hit has been reduced slightly with this version, and I am rerunning that task now, but this isnt properly isolated perf testing so if this is a serious concern we'll need to do proper perf testing.

jayvdb:
master+#61+#68:
real	52m59.855s
user	51m13.372s
sys	0m26.637s

This patch+#61:
real	54m25.811s
user	52m3.561s
sys	0m28.246s

jayvdb:
@sigmavirus24, it looks like an infrastructure failure on Appveyor, due to BitBucket.  I've successfully run the build twice on Appveyor.  Could you restart it, to see if it is working again now on your account.

Also a review would be lovely, especially since you looked at this general problem a while back.
https://bugs.launchpad.net/pyflakes/+bug/1507827

pyflakes/checker.py
```diff
@@ -855,11 +893,18 @@ def handleNode(self, node, parent):
...
+
+        node.depth = self.nodeDepth + 1
         node.parent = parent
-        try:
+
+        if handler is False:
+            return
+
+        if not handler:
             handler = self.getNodeHandler(node.__class__)
```
sigmavirus24: You've moved this outside of the `try`. I suspect it was in there for a reason.

pyflakes/checker.py
```diff
@@ -911,21 +956,22 @@ def ignore(self, node):
         pass
 
     # "stmt" type nodes
-    DELETE = PRINT = FOR = ASYNCFOR = WHILE = IF = WITH = WITHITEM = \
-        ASYNCWITH = ASYNCWITHITEM = RAISE = TRYFINALLY = EXEC = \
-        EXPR = ASSIGN = handleChildren
+    DELETE = PRINT = EXEC = EXPR = RAISE = handleChildrenFlattened
+    ASSIGN = TRYFINALLY = handleChildren
+    FOR = ASYNCFOR = WHILE = IF = WITH = ASYNCWITH = handleChildren
+    WITHITEM = ASYNCWITHITEM = handleChildrenFlattened
```
sigmavirus24: Why not put all of the nodes using `handleChildrenFlattened` together?

pyflakes/checker.py
```diff
@@ -812,9 +812,42 @@ def on_conditional_branch():
...
+                        self.handleChildrenFlattened)
+
+        nodes = [(tree, None)]
+        for node, parent in nodes:
+            # Skip the root of the tree, which has parent None
+            handler = self.getNodeHandler(node.__class__) if parent else False
+            if handler and handler not in _may_flatten:
+                yield node, parent, handler
+            else:
+                nodes[:] += ((child, node)
```
bitglue: What's the `[:]` do here?

jayvdb: List resizing.  https://stackoverflow.com/a/10155987

pyflakes/checker.py
jayvdb: You will need to do better than suspect. :P

pyflakes/checker.py
jayvdb: I have grouped them by type of node in a readible manner, as best I could with this subset of all nodes

I felt the "with-item" nodes should go after the "with" nodes and other branching nodes.

PR #71, opened 2016-05-08
Fix "continue" and "break" being identified as outside a loop when in an "async for" loop

epsy:


jayvdb:
Looks good. Can you amend this PR to be a single squashed commit.

epsy:
Sure thing, done.

ze-phyr-us:
Just ran into this too. Thank you @epsy and looking forward to this being merged and released @jayvdb :-)

jayvdb:
Just noting that that the Python 3.5 [pep-0492](https://www.python.org/dev/peps/pep-0492) support was added in c2086d64 , first released in 1.0.0.

I think this is appropriate for merging and releasing into the 1.2.x series, as it is clearly a bug and has a simple solution.

In any case, master is stable now with the regressions all fixed, so @epsy, if you would like to rebase onto the current master, that will mean it can be merged in fast forward mode, which keeps the commit log cleaner, otherwise a committer would need to amend your patch to do the same.

Maybe also expand the commit message to mention it is specific to Python 3.5 only.

epsy:
OK, done

epsy:
Is there something still preventing this from being merged?

sigmavirus24:
Nope. I think we just lost track of it. :) Thanks @epsy!

djeebus:
Just ran into this issue too, glad to see it's already fixed! Any chance we could get a release?

Thanks!

bitglue:
@djeebus Released 1.3.0.

Thanks @epsy for the contribution, and @jayvdb and @sigmavirus24 for reviewing and merging.

djeebus:
Awesome, thanks all!

PR #72, opened 2016-06-11
Check for duplicate dictionary keys

geokala:


pyflakes/checker.py
```diff
@@ -870,6 +895,33 @@ def ignore(self, node):
...
+        self.handleChildren(node)
+
+    def convert_to_value(self, item):
+        if isinstance(item, ast.Str):
+            return item.s
+        elif isinstance(item, ast.Tuple):
+            return tuple(self.convert_to_value(i) for i in item.elts)
+        elif isinstance(item, ast.Num):
+            return item.n
+        elif isinstance(item, ast.Name):
```
jayvdb: Im pretty sure this fails with instance properties and other attributes of a name.
Could you add it to the test suite , and discard names if I am correct about them.

geokala: Just to make sure I am understanding correctly:
e.g.
```language=python
class Test(object):
    pass
a = Test()
b = Test()
a.something = 'yes'
{
  a.something: 1,
  a.something: 1,
}```
If your assumption is correct, are you expecting this to fail (not detect a duplicate) or fail (traceback)?
I'll give it a test assuming that you meant the above for the moment (good test to have anyway), and if I misunderstood just let me know and I'll deal with it either later today, or tomorrow evening.

Apologies if I'm misunderstanding something trivial- this it the first time I've done much with the AST.

Comment (jayvdb) [inline pyflakes/checker.py]: I expect it to find duplicate 'a' even if different attributes of a are used

Comment (geokala) [inline pyflakes/checker.py]: It didn't seem to find either, as it was an ast.Attribute object. Those are handled now, however.

**Comment by geokala** [comment]
I just added some extra logic for dealing with instance attributes, then realised that I'm also missing logic for dictionary keys (e.g. {aaa['something']: 1, aaa['something']: 1}), and missing tests for the new API elements (e.g. the VariableKey class and the convert_to_values method).

I'll add these over the next couple of days.

**Comment by asmeurer** [comment]
Attribute and getitem lookups and probably venturing too far into the false positive territory. It's possible for a.b and a[b] to return different things on each invocation.

**Comment by geokala** [comment]
True, but is that likely to be deliberately done in more than a tiny fraction of cases?

The only case I can see them returning different things on each invocation are where the attribute is specified as a getter and is set to return something different each time (and similarly for __getitem__).
Am I being blind in considering this extremely rare (and probably a code smell?
(bearing in mind that this shouldn't catch calls to methods such as a.something() being duplicated)

**Comment by jayvdb** [comment]
A good sanity check is running pyflakes, with your new error, over cpython, where it should only occur in tests, and probably not even there.

**Comment by asmeurer** [comment]
Non-pure @property attributes aren't as rare as you would hope.

**Comment by geokala** [comment]
Okay, I'll remove that logic from it shortly then, and get the PR updated. On the bright side, that simplifies the code!

**Comment by geokala** [comment]
Updated.

Based on the existing tests, not testing VariableKey and UnhandledKey would be consistent (though perhaps I should add at least a couple of tests for VariableKey anyway- or should it be trusted that test_other will catch them?).

There also aren't currently tests for convert_to_value. I'm not sure if this is something that there should be tests for or if once again it should be expected that test_other will catch them?

**Comment by jayvdb** [comment]
Your current approach for testing is fine; the utility funcs and classes are getting exercised.

Could you add a test that doesnt raise any errors, with a few valid scenarios.
Also funcs and classes and lambda can be keys.
I dont see a test covering tuple as a key.
Also worth adding a test that `1` and `1.0` are not dup keys, which looks good in your patch but could easily be broken in the future.

Probably also nice to special case `...` as a key that can appear multiple times.

And is there any reason to not extend this test to cover duplicate items in sets?

**Comment by sigmavirus24** [comment]
> And is there any reason to not extend this test to cover duplicate items in sets?

You mean something like:

```py
s = set([
   'foo',
   'bar',
   'foo',
])
```

I think that's less warning worthy than 

```py
d = {
    'foo': 1,
    'bar': 2,
    'foo': 3,
}
```

Given sets have defined behaviour and there is undefined behaviour in the dictionary case.

geokala:
Agreed- if I initialise a set with duplicate values I expect it to discard the duplicates. If I do it with a dictionary then I get unexpected effects (until I track down the issue by hand).

geokala:
@jayvdb: I'll probably address your other comments this evening, should have an update in then.

sigmavirus24:
> Probably also nice to special case ... as a key that can appear multiple times.

Wait, why? That doesn't work as `...` is a singleton:

```py
>>> {...: 'foo'}
{Ellipsis: 'foo'}
>>> {...: 'foo', ...: 'bar'}
{Ellipsis: 'bar'}
```

geokala:
Out of interest, what is the utility of allowing ... to appear as a dictionary key multiple times?

I can't think of a situation in which it'd be useful off the top of my head.

Happy to special-case it, just a bit perplexed as to why anyone would do it in the first place.

sigmavirus24:
> Out of interest, what is the utility of allowing ... to appear as a dictionary key multiple times?

I agree. I'm not sure I see the value. I'm open to an explanation for *why* it should be special-cased though.

geokala:
> Could you add a test that doesnt raise any errors, with a few valid scenarios.
I've added a few tests that don't raise errors, and a test to see if we pick up duplicate keys in a literal dictionary inside a lambda.

> Also funcs and classes and lambda can be keys.
Functions and classes are ast.Names.
Lambdas, unless assigned to a variable (which are already caught when duplicated) don't 'count' as duplicates- see:
```{
   lambda x: None: 1,
   lambda x: None: 1,
}```
This actually gives a dictionary with two items.

> I dont see a test covering tuple as a key.
https://github.com/pyflakes/pyflakes/pull/72/files#diff-d18f55ece21d21fb172322f7db700d6aR31

> Also worth adding a test that 1 and 1.0 are not dup keys, which looks good in your patch but could easily be broken in the future.
Those are actually duplicates (tested in python2 and python2 with {1.0: 1, 1:1} ) and are picked up correctly.

**Comment by jayvdb** [comment]
wrt to `...`, it is a legal expression in py3.  Its use as a placeholder for code is accepted.
https://mail.python.org/pipermail/python-3000/2008-January/011793.html
Using `...` is clear intent by the author (to do the wrong thing) , but it is valid syntax and Python doesnt have any attached meaning for it, so IMO pyflakes shouldnt declare invalid syntax if it is used as a placeholder twice in the same dict:
```
a = {
    1: 1,
    2: 2,
    3: 3,
    ...: ...,
    'a': 'a',
    'b': 'b',
    'c': 'c',
    ...: ...,
}
```
There is only a very small chance we're helping anyone with that being an error.  If someone is using `...` in a dict, they shouldnt be expecting Python syntax checkers to be assisting them.  It cant be a typo as far as I can see, as there is no valid syntax which is lexically similar to `...`.  They intended to use it, their intent indicates the code is not production code, it is valid to use it, so why error?

If pyflakes is going to error when `...` is used in that dict, we should error whenever it appears in the dict (not just the first time), or possibly anywhere in code except in slices.  But that is assuming a meaning for `...`, and shouldnt be done by pyflakes IMO (but would be very useful , similar to flake8_print and similar which check that code is ready for promoting).

This is in contrast to the other similar constant `NotImplemented`, which has no placeholder value and it is reasonable to have it appear in a dict of real production code, and only one dict entry would be valid.

**Comment by jayvdb** [comment]
@sigmavirus24 , no I do not mean `s = set(['foo', 'bar', 'foo',])` .  That is a list display in a `set` constructor invocation.  I am of course referring to set displays.  i.e. `s = {'a', 'b', 'c', }`.  That should be the natural assumption in technical assumptions on pyflakes, as we work on the ast level.

> Given sets have defined behaviour and there is undefined behaviour in the dictionary case.

What is the undefined behaviour in the dictionary case?
I believe any ambiguity was removed almost [9 years ago](https://github.com/python/cpython/commit/516d2cf2cb07fba20beed26372fece867c370393#diff-10e0813310f7198e9e0916db3e093154R254 ) in the Python 2.5.2 documentation. Am I missing something?

https://docs.python.org/3/reference/expressions.html#dictionary-displays
(and https://docs.python.org/2/reference/expressions.html#dictionary-displays)

File: pyflakes/test/test_other.py [hunk 2]
```diff
@@ -13,6 +13,97 @@ class Test(TestCase):
...
+            expected = m.DuplicateDictionaryKey
+        self.flakes("{None: 1, None: 1}", expected)
+
+    def test_duplicate_variable_keys(self):
+        self.flakes('''
+        a = 1
+        {a: 2, a: 3}
+        ''', m.DuplicateVariableDictionaryKey)
+
+    def test_duplicate_key_float_and_int(self):
```
jayvdb: This one really surprised me, as they are stored differently
```
>>> {1.0: 'a'}
{1.0: 'a'}
>>> {1: 'a'}
{1: 'a'}
```
The same occurs for sets.
```
>>> {1, 1.0}
set([1.0])
```

But it is documented behaviour (https://github.com/python/cpython/blame/master/Doc/library/stdtypes.rst#L4008) . (Probably would be useful for the Python reference on dict/set displays to mention it also)

Anyway, it will be very helpful for pyflakes to be catching this oddity.  IMO it justifies any 'false positives' that this improvement might have.

IMO this test deserves a docstring ;-)

pyflakes/test/test_other.py
```diff
@@ -13,6 +13,97 @@ class Test(TestCase):
...
+
+    def test_no_duplicate_key_errors_bool_or_none(self):
+        self.flakes("{True: 1, None: 2, False: 1}")
+
+    def test_no_duplicate_key_errors_ints(self):
+        self.flakes('''
+        {1: 1, 2: 1}
+        ''')
+
+    def test_no_duplicate_key_errors_vars(self):
```
jayvdb: IMO this is a TODO ; we have all the information we need to detect the duplicate key.

pyflakes/test/test_other.py
```diff
@@ -13,6 +13,97 @@ class Test(TestCase):
...
+        self.flakes('''
+        test = 'yes'
+        rest = 'yes'
+        {test: 1, rest: 2}
+        ''')
+
+    def test_no_duplicate_key_errors_tuples(self):
+        self.flakes('''
+        {(0,1): 1, (0,2): 1}
```
jayvdb: Worth also adding 1 vs 1.0, as in tuples they _are_ different.
```
>>> {(1,1): 1, (1.0 ,2): 1}
{(1.0, 2): 1, (1, 1): 1}
```

pyflakes/messages.py
```diff
@@ -116,6 +116,22 @@ def __init__(self, filename, loc, name):
         self.message_args = (name,)
 
 
+class DuplicateDictionaryKey(Message):
+    message = 'dictionary key %s duplicated'
```
jayvdb: I have a slight preference for "duplicated" being "repeated" everywhere, or at least in these messages, as that is how cpython describes it.

```bash
$ python -c "def foo(a, b): pass; foo(a=1, a=2)"
  File "<string>", line 1
SyntaxError: keyword argument repeated
```

But also "repeated" is much more accessible for ESL people.

"This word was duplicated." has a Flesch reading ease score of 54.7 (and is roughly ~US school grade 10) , whereas "This word was repeated." is 75.9 (~ grade 6)

pyflakes/test/test_other.py
```diff
@@ -13,6 +13,97 @@ class Test(TestCase):
...
+
+    def test_duplicate_keys_ints(self):
+        self.flakes("{1: 1, 1: 1}", m.DuplicateDictionaryKey)
+
+    def test_duplicate_keys_bools(self):
+        if version_info < (3, 4):
+            expected = m.DuplicateVariableDictionaryKey
+        else:
+            expected = m.DuplicateDictionaryKey
+        self.flakes("{True: 1, True: 1}", expected)
```
jayvdb: IMO pyflakes should hide the fact that the ast is slightly different for these constants (True, False, None) for 3.4+ , so that the error message can be consistent between Python versions.  Our users dont consider `True` to be a variable on Python < 3.4.

pyflakes/messages.py
sigmavirus24: That makes perfect sense. Thanks for being conscious of that @jayvdb. @geokala can you fix this up?

geokala: Sure, I'll be looking at all comments and updating the PR/comments (hopefully) appropriately this evening.

pyflakes/test/test_other.py
geokala: I considered this, but I'm unsure. In most cases I would expect that variables will be being dynamically set, such that adding detection that actually looks up values where possible is a lot more effort than the reward it gains.

However, I suppose in cases of constants defined in the code it may well have some value so agreed that TODO is valid for this.

geokala:
> duplicated -> repeated

Done in the error messages, but I've left the (message) class names as they are, as they mesh with the 'DuplicateArgument'. Can change if that's preferred, but as it's under the hood it's probably better to have similar naming to similarish issues?

> 1.0 vs 1 in tuples

I see the same behaviour there as in dict keys (py2 and py3):
{(1,0): 1, (1.0, 0): 2} -> {(1, 0): 2}

I've added an extra test anyway to check for:
{(1,0): 1, (1, 1): 2} (to confirm that we don't get a clash when just one element of a tuple is the same
and
{(1,0): 1, (1.0, 0): 2 (to make it clear that this behaviour is expected and to alert us if it changes in future versions of python... not that that's likely?)

> Hiding non-constant nature of True,False,None in py<3.4

Done.

pyflakes/test/test_other.py
geokala: Spoken like someone who has never had a developer try to ban them from Booleans due to (deliberately) redefining False in a dojo environment. *smiles*

pyflakes/test/test_other.py
jayvdb: Hmm. You had different results? I was probably using 3.5.1 here (im effectively offline atm, in an angkot)

pyflakes/test/test_other.py
jayvdb: Agreed the benefits are small. It would be an easy project for a uni student or Google Code-in as it is very self contained.

pyflakes/test/test_other.py
geokala: Travis gives the same results for 3.5 as well (and nightly, which I'm guessing is at least 3.5.1 equivalent).

I think your example may be giving you different results because your second element is different (I did include a test that we don't incorrectly claim equality based on only the first element after seeing that though- seemed like something worth checking).

pyflakes/test/test_other.py
geokala: Docstring added.

geokala:
Okay, I think I've addressed all of the comments again. Is there anything I've missed (or any new thoughts)?

jayvdb:
Waiting for @sigmavirus24 to explain why set views shouldnt be included.

geokala:
My own two pence on set views:
If I put something twice (or more) into a dictionary it's probably because I've considered each valid at one point, and now one of them is going to 'win' and possibly cause hard to understand bugs.

If I put something twice (or more) into a set view, I want a set of the results with duplicates removed, so while it may /look/ a little smelly, it's not actually going to be causing 'interesting' runtime bugs- the set will still have a copy of that element that I added to it.

jayvdb:
In that is the rationale for not including set views, we should only error if the repeated key has a different value.

sigmavirus24:
> If that is the rationale for not including set views, we should only error if the repeated key has a different value.

That seems fair to me.

pyflakes/test/test_other.py
sigmavirus24: Nightly should be 3.6.0 pre-releases.

geokala:
That would seem to fit with the aim of pyflakes ("never complain about style").

geokala:
Update in place with a couple more tests and comments for the new behaviour (only alerting when dictionary keys are repeated with differing values).

I have not changed the messages at this time. I think they still tell the right 'story', so the main reason to change them in my view would be if we were concerned about someone wondering why a repeated key+value pair didn't cause the alert. However, that's probably a style issue rather than a syntax one.

jayvdb:
I would prefer the message mentions that the values differ, to avoid bug reports for the same key & value, or requesting the same feature for sets, but happy to accept the current shorter messages.

pyflakes/checker.py
geokala: (note as this still appears in its own thread on the PR: These are no longer handled- see remaining comments)

geokala:
Okay, I've made a change so that both messages now mention the differing values:
```test.py:3: dictionary key yes repeated with different values```

**Comment by jayvdb** [comment]
In general, yes, testing the message strings is overcooking the tests, at least until someone improves the test framework so checking the message string is able to be done in a way that doesnt hurt readability.

I've done it once or twice. See `test_unusedImport_relative`, but also see the first class in test_imports where I test the source of the varying content in the messages. But that was a more complex scenario than your change here. IMO one check of the message string would be more than enough, and isnt necessary here.

File: pyflakes/checker.py [hunk 7]
```diff
@@ -870,6 +895,63 @@ def ignore(self, node):
...
+            result = VariableKey(item=item)
+            constants_lookup = {
```
jayvdb: I'm sure we will have other uses for this lookup, so it could be a module level constant.

pyflakes/checker.py
```diff
@@ -870,6 +895,63 @@ def ignore(self, node):
...
     COMPREHENSION = KEYWORD = FORMATTEDVALUE = handleChildren
 
+    def DICT(self, node):
+        # Complain if there are duplicate keys with different values
+        # If they have the same value it's not going to cause potentially
+        # unexpected behaviour so we'll not complain.
+        keys = [
+            self.convert_to_value(key) for key in node.keys
+        ]
+        values = [
```
jayvdb: It would be nicer for perf and memory if we didnt need to do `convert_to_value` (which also creates new objects) for every dict value: only values for repeated keys should be analysed in this logic.

jayvdb:
Another fun one... b'a' and u'a' have different values in `ast.Str.s` on py2 at least, but are repeated keys in dict views.

pyflakes/checker.py
jayvdb: (thx, I wish there was a way to force a pr thread to be marked as done. I find github to be very confusing for in-depth code reviewing with multiple incremental revisions being made. The result seems to be that committers tend to merge early and do fixups after, creating churn and often bugs.)

pyflakes/checker.py
geokala: Okay, next push that will be changed to do it when it checks the values of each key later (only on ones that have the key).

geokala:
As for {b'a': 1, u'a': 2}, those are indeed not duplicated in python3, so I'll put a test in to confirm the behaviour.

I think using both of those in python3 would be a really strong code smell in most cases, but... that's not something pyflakes should complain about so I won't worry.

Either way, it will now handle bytes as well so... positive result!

pyflakes/checker.py
geokala: Moved to module level

pyflakes/messages.py
```diff
@@ -116,6 +116,22 @@ def __init__(self, filename, loc, name):
...
 
+class DuplicateDictionaryKey(Message):
+    message = 'dictionary key %s repeated with different values'
+
+    def __init__(self, filename, loc, key):
+        Message.__init__(self, filename, loc)
+        self.message_args = (key,)
+
+
+class DuplicateVariableDictionaryKey(Message):
```
jayvdb: I find this name to be confusing, as my brain reads it as being a "duplicate variable something ..."; the words 'duplicate' and 'key' are too far removed in the class name, and it doesnt help that this is now used to report conflicting values for the same key.

MultipleValueRepeatedLiteralKey and 
MultipleValueRepeatedNameKey ?

Im wondering whether we need two message classes; the word 'variable' could be added to the outputted message by VariableKey.__str__.

pyflakes/test/test_other.py
```diff
@@ -13,6 +13,130 @@ class Test(TestCase):
     def test_duplicateArgs(self):
         self.flakes('def fu(bar, bar): pass', m.DuplicateArgument)
 
+    def test_duplicate_keys(self):
+        self.flakes("{'yes': 1, 'yes': 2}", m.DuplicateDictionaryKey)
+
+    @skipIf(version_info >= (3,),
+            "bytes and strings with same 'value' are not equal in python3")
+    def test_duplicate_keys_bytes_vs_unicode(self):
```
jayvdb: It would be helpful to do a similar py2 check with `{1: b'a', 1: u'a'}` .

It also wouldnt hurt adding py3 checks, confirming that it treats b'a' and u'a' as non-equal, even though that is one of the defining characteristics of py3. It would be more for regression testing to prevent someone royally screwing up in the future if they modified your algorithm.

pyflakes/checker.py
```diff
@@ -870,6 +922,38 @@ def ignore(self, node):
     # additional node types
     COMPREHENSION = KEYWORD = FORMATTEDVALUE = handleChildren
 
+    def DICT(self, node):
+        # Complain if there are duplicate keys with different values
+        # If they have the same value it's not going to cause potentially
+        # unexpected behaviour so we'll not complain.
```
jayvdb: Useful to special case `{}` by bailing out before building the following array?

jayvdb:
Maybe create `test_dict`? It seems there are enough corner cases, and some areas of expansion not yet explored (e.g. looking at variable values), that this deserves its own test module.

pyflakes/checker.py
geokala: From a very quick and dirty test (timing 10000 runs of a func which does result = [str(key) for key in item.keys()] with no gating, a test for == {} and a test for len(thing) == 0), it seems like the list comprehension finishes just as fast as the gated versions (possibly faster actually, though that may be an artefact of the way the test was constructed).

Probably not worth it given that it gives no obvious gain in efficiency anyway.

pyflakes/test/test_other.py
geokala: Added, will be in next update.

pyflakes/messages.py
geokala: Agreed that the name probably needs improving. I'll do something with it momentarily.

Having two different classes for this allows the tests to be improved somewhat- so that we can tell a user that they have used duplicates of the string 'myvariable' vs the variable myvariable.

They're currently:
Duplicate<x>Variable: Any duplicated keys which are variables, including classes and functions, but NOT including any builtins (even in python2, special cased True,False,None).
This particular special case was introduced in regards to an earlier comment which (sensibly, I think?) pointed out that most people do not consider True, False, and None to be variables in python2 (even though they sort of are).

Duplicate<x>: Any duplicated keys which are builtins or other literals.

For the moment I'll just rename these to something stating Variable vs Literal and hopefully removing the disconnect. They can always be improved further as desired.

geokala:
Okay, I think this addresses everything (though I've got a sneaking suspicion I missed something but couldn't find anything obvious when I looked).

For the unicode strings I've skipped the step on 3.2. I could remove the unicode tag from the string (as it will be unicode anyway), but then it'll make it slightly less clear precisely what the test is doing so part of me feels like skipping that test for python3.2 is the lesser evil, but I'm not certain either way on that.

geokala:
Is there anything else I can do to improve this and help get it to the point where it can be merged?

jayvdb:
My guess is we've caught everything.
Ill do some testing on it today.

sigmavirus24:
So I did some light testing on this today and it seems to work just fine. I'm in favor of merging it and will do so tonight if no one objects.

I tested against openstack/nova to see if there was a performance difference between the two versions (as the project is very large) and there was nothing statistically significant in the difference between 1.2.3 on PyPI and what's here in the repository.

jayvdb:
Sorry I havent finished testing this, yet.

So far I have only found one significant problem.

It is reporting the line number of the beginning of the fict for all instances of the repeated key.

```bash
cpython$ pyflakes . 2>&1 | grep 'dictionary key'
./Lib/mimetypes.py:402: dictionary key .cdf repeated with different values
./Lib/mimetypes.py:402: dictionary key .xls repeated with different values
./Lib/encodings/mac_arabic.py:441: dictionary key 32 repeated with different values
./Lib/encodings/mac_arabic.py:441: dictionary key 33 repeated with different values
./Lib/encodings/mac_arabic.py:441: dictionary key 34 repeated with different values
./Lib/encodings/mac_arabic.py:441: dictionary key 35 repeated with different values
./Lib/encodings/mac_arabic.py:441: dictionary key 36 repeated with different values
./Lib/encodings/mac_arabic.py:441: dictionary key 38 repeated with different values
./Lib/encodings/mac_arabic.py:441: dictionary key 39 repeated with different values
./Lib/encodings/mac_arabic.py:441: dictionary key 40 repeated with different values
./Lib/encodings/mac_arabic.py:441: dictionary key 41 repeated with different values
./Lib/encodings/mac_arabic.py:441: dictionary key 42 repeated with different values
./Lib/encodings/mac_arabic.py:441: dictionary key 43 repeated with different values
./Lib/encodings/mac_arabic.py:441: dictionary key 45 repeated with different values
./Lib/encodings/mac_arabic.py:441: dictionary key 46 repeated with different values
./Lib/encodings/mac_arabic.py:441: dictionary key 47 repeated with different values
./Lib/encodings/mac_arabic.py:441: dictionary key 58 repeated with different values
./Lib/encodings/mac_arabic.py:441: dictionary key 60 repeated with different values
./Lib/encodings/mac_arabic.py:441: dictionary key 61 repeated with different values
./Lib/encodings/mac_arabic.py:441: dictionary key 62 repeated with different values
./Lib/encodings/mac_arabic.py:441: dictionary key 91 repeated with different values
./Lib/encodings/mac_arabic.py:441: dictionary key 92 repeated with different values
./Lib/encodings/mac_arabic.py:441: dictionary key 93 repeated with different values
./Lib/encodings/mac_arabic.py:441: dictionary key 94 repeated with different values
./Lib/encodings/mac_arabic.py:441: dictionary key 95 repeated with different values
./Lib/encodings/mac_arabic.py:441: dictionary key 123 repeated with different values
./Lib/encodings/mac_arabic.py:441: dictionary key 124 repeated with different values
./Lib/encodings/mac_arabic.py:441: dictionary key 125 repeated with different values
./Lib/test/test_mailbox.py:2236: dictionary key Received repeated with different values
./Lib/test/test_mailbox.py:2236: dictionary key Delivered-To repeated with different values
```

`./Lib/mimetypes.py` is fairly easy to spot
```python
    types_map = {  # line 402
        ...
        # Duplicates :(
        '.cdf'    : 'application/x-cdf',
        '.cdf'    : 'application/x-netcdf',
        ...
    }
```

`./Lib/test/test_mailbox.py` is: less easy to see:
```python
_sample_headers = {  # line 2236
    "Return-Path":"<gkj@gregorykjohnson.com>",
    "X-Original-To":"gkj+person@localhost",
    "Delivered-To":"gkj+person@localhost",
    "Received":"""from localhost (localhost [127.0.0.1])
        by andy.gregorykjohnson.com (Postfix) with ESMTP id 356ED9DD17
        for <gkj+person@localhost>; Wed, 13 Jul 2005 17:23:16 -0400 (EDT)""",
    "Delivered-To":"gkj@sundance.gregorykjohnson.com",
    "Received":"""from localhost [127.0.0.1]
        by localhost with POP3 (fetchmail-6.2.5)
        for gkj+person@localhost (single-drop); Wed, 13 Jul 2005 17:23:16 -0400 (EDT)""",
    "Received":"""from andy.gregorykjohnson.com (andy.gregorykjohnson.com [64.32.235.228])
        by sundance.gregorykjohnson.com (Postfix) with ESMTP id 5B056316746
        for <gkj@gregorykjohnson.com>; Wed, 13 Jul 2005 17:23:11 -0400 (EDT)""",
    "Received":"""by andy.gregorykjohnson.com (Postfix, from userid 1000)
        id 490CD9DD17; Wed, 13 Jul 2005 17:23:11 -0400 (EDT)""",
    "Date":"Wed, 13 Jul 2005 17:23:11 -0400",
    "From":""""Gregory K. Johnson" <gkj@gregorykjohnson.com>""",
    "To":"gkj@gregorykjohnson.com",
    "Subject":"Sample message",
    "Mime-Version":"1.0",
    "Content-Type":"""multipart/mixed; boundary="NMuMz9nt05w80d4+\"""",
    "Content-Disposition":"inline",
    "User-Agent": "Mutt/1.5.9i" }
```

`./Lib/encodings/mac_arabic.py` is even harder, as the source and error report refer to two different numbers.

```python

encoding_map = {  # line 441
    ...
    0x0020: 0x0020,     #  SPACE, left-right
    0x0020: 0x00a0,     #  SPACE, right-left
    0x0021: 0x0021,     #  EXCLAMATION MARK, left-right
    0x0021: 0x00a1,     #  EXCLAMATION MARK, right-left
    0x0022: 0x0022,     #  QUOTATION MARK, left-right
    0x0022: 0x00a2,     #  QUOTATION MARK, right-left
    0x0023: 0x0023,     #  NUMBER SIGN, left-right
    0x0023: 0x00a3,     #  NUMBER SIGN, right-left
    0x0024: 0x0024,     #  DOLLAR SIGN, left-right
    0x0024: 0x00a4,     #  DOLLAR SIGN, right-left
    0x0025: 0x0025,     #  PERCENT SIGN, left-right
    0x0026: 0x0026,     #  AMPERSAND, left-right
    0x0026: 0x00a6,     #  AMPERSAND, right-left
    0x0027: 0x0027,     #  APOSTROPHE, left-right
    0x0027: 0x00a7,     #  APOSTROPHE, right-left
    0x0028: 0x0028,     #  LEFT PARENTHESIS, left-right
    0x0028: 0x00a8,     #  LEFT PARENTHESIS, right-left
    0x0029: 0x0029,     #  RIGHT PARENTHESIS, left-right
    0x0029: 0x00a9,     #  RIGHT PARENTHESIS, right-left
    0x002a: 0x002a,     #  ASTERISK, left-right
    0x002a: 0x00aa,     #  ASTERISK, right-left
    0x002b: 0x002b,     #  PLUS SIGN, left-right
    0x002b: 0x00ab,     #  PLUS SIGN, right-left
    0x002c: 0x002c,     #  COMMA, left-right; in Arabic-script context, displayed as 0x066C ARABIC THOUSANDS SEPARATOR
    0x002d: 0x002d,     #  HYPHEN-MINUS, left-right
    0x002d: 0x00ad,     #  HYPHEN-MINUS, right-left
    0x002e: 0x002e,     #  FULL STOP, left-right; in Arabic-script context, displayed as 0x066B ARABIC DECIMAL SEPARATOR
    0x002e: 0x00ae,     #  FULL STOP, right-left
    0x002f: 0x002f,     #  SOLIDUS, left-right
    0x002f: 0x00af,     #  SOLIDUS, right-left
    ...
}
```

It should report the line number of the offending keys, string keys should be in quotes, and (less important) other keys types should report the same value as in the source .

sigmavirus24:
So that's definitely plausible, I inserted a break in the code and did a little exploring:

```py
# t.py
a = {
    'a': 'b',
    'a': 'c',
}
```

```
(Pdb) l
935  	        keys = [
936  	            convert_to_value(key) for key in node.keys
937  	        ]
938  	        import pdb
939  	        pdb.set_trace()
940  ->	        for key in set(keys):
941  	            if keys.count(key) > 1:
942  	                candidate_values = []
943  	                for candidate in range(len(keys)):
944  	                    if key == keys[candidate]:
945  	                        candidate_values.append(
(Pdb) p node
<_ast.Dict object at 0x10ad3d510>
(Pdb) p dir(node)
['__class__', '__delattr__', '__dict__', '__doc__', '__format__', '__getattribute__', '__hash__', '__init__', '__module__', '__new__', '__reduce__', '__reduce_ex__', '__repr__', '__setattr__', '__sizeof__', '__str__', '__subclasshook__', '__weakref__', '_attributes', '_fields', 'col_offset', 'depth', 'keys', 'lineno', 'parent', 'values']
(Pdb) p node.lineno
1
(Pdb) p node.values
[<_ast.Str object at 0x10ad3d5d0>, <_ast.Str object at 0x10ad3d610>]
(Pdb) p node.values[0].lineno
2
(Pdb) p node.values[1].lineno
3
```

So if we don't convert everything to its value, we can probably grab the line number rather easily.

Alternatively, we can create a class that all values can be shoved into as a generic and that can keep this information. @geokala are you up for that or do you want @jayvdb or myself to take over?

geokala:
Modification made to grab it (hopefully that answers your question *smiles*).

It now gives the line number as the /first/ occurrence of the key definition. It would probably be possible to make it give a later one- perhaps the last one- if I play with reversing the lists.

I think that's what you intended?

I wouldn't expect it to have a big impact on performance- it does an extra index check and then lookup on a list by key only in the case that there is a duplicate. Where no duplicate is found there should be zero impact.

sigmavirus24:
So I don't think the first one is the one we want to flag/report unless we want to change the message to. I think we'd want to flag *each one* after the first.

If reporting on the first line is easiest, then I'd like to suggest we say "dictionary key <key> repeated with different values on lines ..."

Also, I just noticed that the dictionary key is using `%s` instead of `%r` for formatting. For string keys, it could be confusing if you don't see `''`s around the string. Ints will be represented the same way, and tuples should work the same way as well.

pyflakes/checker.py
```diff
@@ -870,6 +922,41 @@ def ignore(self, node):
...
+        keys = [
+            convert_to_value(key) for key in node.keys
+        ]
+        for key in set(keys):
+            if keys.count(key) > 1:
```
lamby: Feels like this would be cleaner (eg. less cyclomatic complexity) if written as

```python
if keys.count(key) == 0:
    continue

candidate_values = []
# [..]
```

pyflakes/checker.py
```diff
@@ -870,6 +922,41 @@ def ignore(self, node):
...
+                candidate_values = []
+                for candidate in range(len(keys)):
+                    if key == keys[candidate]:
+                        candidate_values.append(
+                            convert_to_value(node.values[candidate])
+                        )
+
+                for value in set(candidate_values):
+                    if candidate_values.count(value) == 1:
```
lamby: (Same here re. inverting the conditional and using `continue`)

pyflakes/checker.py
sigmavirus24: The inversion of that isn't `if keys.count(key) == 0` though. It's `if keys.count(key) <= 1`.

lamby: Haha. Good spot, thanks!

jayvdb:
Ideally each reported error has the line number where the value is. Other tools use this line number to show the source line for that specific error.

I agree some continues would help reduce the indents.

jayvdb:
Another strange case (on 2b86b4e):

```bash
distlib$ pyflakes . 2>&1 | grep 'dictionary key'
./tests/test_metadata.py:146: dictionary key author repeated with different values
./tests/test_metadata.py:169: dictionary key author repeated with different values
```

This error stood out because there should never be a case where only one error appears - to be repeated there must be two errors (and currently they are reported with the same line number)

Sure enough, they are two different dicts that look like this:

```python
        metadata = LegacyMetadata(mapping={'author': 'Mister Café',  # line 146
                                     'name': 'my.project',
                                     'author': 'Café Junior',
                                     'summary': 'Café torréfié',
                                     'description': 'Héhéhé',
                                     'keywords': ['café', 'coffee']
                                  })
```

Somehow it is only reporting line 146, and not reporting the dup on 148.

geokala:
@jayvdb: Sorry- I must've misunderstood. Previous behaviour (up until the change I just pushed) was to raise only one complaint per set of duplicated keys with differing values (e.g. { 1: 1, 1: 2 } would raise one complaint). I've modified it now so it should behave as you expect.

@lamby: I've also put in the change you and @sigmavirus24 suggested above (well, you suggested, with addendum, etc, etc). It could actually be changed to == 1 (it can't ever be zero for the comparison it's doing), but I'm not sure if that gives any real improvement.

pyflakes/checker.py
```diff
@@ -870,6 +922,49 @@ def ignore(self, node):
...
+                    )
+
+            # Determine whether we have different values for the repeats
+            differing_values = False
+            for value in set(candidate_values):
+                if candidate_values.count(value) == 1:
+                    differing_values = True
+                    break
+
+            if differing_values:
```
lamby: another one for inverting the logic and ``continue``?

lamby:
> but i'm not sure if that gives any real improvement.

Really? Reducing indentation makes it far, far easier to follow IMHO

pyflakes/checker.py
jayvdb: The whole differing_values section can be reduced to a single `if [any|all](): continue` , which reduces the complexity of the method by ~16%

jayvdb:
@geokala , thanks for including reporting of all the lines with duplicates, and adding the correct line numbers.

Could you please use `%r` for the literals, which will add quotes around the strings, which is quite necessary if the string contains a space, or worse if the string contains newline characters.

I'd also really like the reported literal to be same as appears in the source, rather than key `0x0020` appearing in the report output as 32.  If I see a literal in a error report, I expect to be able to grep the source file for that literal.  The perfect UI IMO would emit `... "0x0020" (32) ...`, as then it is grepable, and two different ways of writing the same number are shown as the same number.
To do this, it looks like we need to use `dict_node.keys[i].col_offset`, and parse for the end of the key (and give up if we somehow reach the corresponding `dict_node.values[i].col_offset` ?).  Or is there another way I'm not seeing?  Maybe this should be a followup patch.

re complexity/indents/etc:

```
$ python -m mccabe pyflakes/checker.py | grep DICT
("925:1: 'Checker.DICT'", 12)
```

That makes it the fourth most complex method, and the second most complex node handler method.  Which is not representative of the actual complexity of the task.  I've noted above one way to reduce that to 10, and I suspect one or two more generators/comprehensions could be added to make the complex parts can be easy to load into the brain as a single chunk.

geokala:
@lamby: Apologies, I was unclear in my meaning- I meant that changing it to == 1, rather than <=1, while filling the same function here (and functioning the same way currently), probably doesn't gain us anything (and may open up future bug possibilities if the logic is slightly changed). The original change, yes, no complaints. -smiles-

@jayvdb: Sorry, I must've missed that one- I recall intending to do it. Either way, it should be good now.
I think the reported literal issue for hex (and similar) is going to be worthy of a followup patch, yes. As you suggest, it is somewhat non-trivial (unless someone knows a way).
I'll do another pass now to try to get the complexity below 10.

geokala:
@jayvdb: Change made per your suggestion of any(x) for differing_values.
Complexity now 10.

I'm not sure if it can be reasonably reduced in complexity now (in a way that doesn't sacrifice readability), but that may be down to my skill level and/or perceptions.

jayvdb:
Take a look at https://github.com/jayvdb/pyflakes/commit/3b29b50638d9b77372727160b5d13bcfd6420228 for some ideas on how to cut complexity to 5.

sigmavirus24:
@jayvdb I don't know of a specific guideline for what the complexity of the code should be. If there is one, we should add it to Travis CI and document it for future contributors. At this point, playing complexity golf with @geokala seems rude. If you have the complexity that low in your own fork, feel free to submit it as a pull request after this one. I think @geokala has been more than patient with us here.

sigmavirus24:
Since we're just golfing over complexity, I'm going to run this over some more code-bases today and if all seems good, I'm going to merge this.

If there are reviews other than complexity golfing, please do leave them. I think this pull request has lived for quite long enough. Other style nits should be submitted by people after merge.

geokala:
@sigmavirus24 Sounds fair. I think @jayvdb's changes were quite elegant (I should've guessed collections would have something to simplify what I was trying to do there), but if you're happy running over this and then determining complexity intent later that's cool.

Also, yes, I do think determining a complexity guideline and testing for it (though maybe with an easy and obvious way to add exceptions) might be a good idea. I'm fairly sure flake8 supports this natively (as I think I actually turned it on at the start of a project in a previous workplace).

Either way, if there are any other issues, I'm happy to address them!

jayvdb:
My review comments are not just style nits.  They are also performance improvements.

sigmavirus24:
@jayvdb there is one review comment here unrelated to complexity that @geokala has kindly addressed. Unless there are more comments that are not being displayed by GitHub, all your existing comments relate to complexity.

sigmavirus24:
Also @geokala if you want to adopt @jayvdb's changes, go ahead. IMO, you don't have to keep playing complexity golf though. You were asked to make it no higher than 10 and did so.

geokala:
I'm happy to adopt them, though it'll probably be late tonight before I get to them.

jayvdb:
Ian, I think you have misread my comment earlier about '10' and if you look briefly at my diff you will see it is performance improvements implementing the additional generators/comprehensions I had indicated earlier would help get the complexity below 10, down to a reasonable level. The current version is twice reimplementing a Counter, using list scans. It is best practise when doing code reviews to not worry about those kind of review comments until the patch is functionally ok. I put a diff up because that was easier than trying to explain how to fix it, which would have been a longer and more aggravating process for everyone involved, as it involved rearranging the code significantly. But the code isnt mine, and I wouldntvwant

jayvdb:
While looking at pyflakes output, it occurred to me that this feature is basically identical to the existing `RedefinedWhileUnused` which emits `"filename.py:875: redefinition of unused 'xxx' from line 863` , except it is a prior dict key value that is unused rather than a variable value.

If there are any complaints after this is release that this is another case of `pyflakes` violating its contract with users and producing an error on valid code, `RedefinedWhileUnused` is a simple way to demonstrate that there is precedent for it.

Also, if we wanted to only emit only one error line for a pair of repeated keys, using the same suffix of `RedefinedWhileUnused` being `... from line x` referring at the prior & different use of the key would be the way to do it.  Note: I'm not asking for, or advocating for, it to be modified; im just saying that would be reasonable IMO.
The current approach of emitting two messages for a pair of repeated keys is a bit verbose, but it is my preference as it works better with the `flake8 --show-source` as both lines with the different values are shown.

geokala:
Hmm... collections has no Counter in 2.6... @jayvdb, do we have a good alternative other than using a backport?

jayvdb:
Sure.  We only need a dict with the value incrementing, as we dont need any of the fancy Counter features.
Best to build it as a fallback class for when the real Counter doesnt exist, but it could also be a normal (default)dict incrementing values using a for loop.

geokala:
@jayvdb: Okay, I've put in a counter method and just used it, rather than putting extra complexity in to determine whether to use it.

@sigmavirus24 (copying you on this if you want to give it another look and/or if you're both happy to merge?)

jayvdb:
Nice work.  Thank you so much for sticking with it.

geokala:
No worries- thanks to all of you for your comments!

PR #73, opened 2016-06-14
Ignore underscore redefinition

eight04:
https://github.com/kevinw/pyflakes/issues/30

### DIFF
--- a/pyflakes/test/test_other.py
+++ b/pyflakes/test/test_other.py
@@ -161,6 +161,15 @@ class Test(TestCase):
         def _(): pass
         ''')
 
+    def test_redefinedUnderscoreImportation(self):
+        """
+        Test that shadowing an underscore importation raises a warning.
+        """
+        self.flakes('''
+        from .i18n import _
+        def _(): pass
+        ''', m.RedefinedWhileUnused)
+
     def test_redefinedClassFunction(self):
         """
         Test that shadowing a function definition in a class suite with another
--- a/pyflakes/test/test_other.py
+++ b/pyflakes/test/test_other.py
@@ -151,6 +151,16 @@ class Test(TestCase):
         def a(): pass
         ''', m.RedefinedWhileUnused)
 
+    def test_redefinedUnderscoreFunction(self):
+        """
+        Test that shadowing a function definition named with underscore doesn't
+        raise anything.
+        """
+        self.flakes('''
+        def _(): pass
+        def _(): pass
+        ''')
+
     def test_redefinedClassFunction(self):
         """
         Test that shadowing a function definition in a class suite with another
--- a/pyflakes/checker.py
+++ b/pyflakes/checker.py
@@ -170,6 +170,9 @@ class Definition(Binding):
     A binding that defines a function or a class.
     """
 
+    def redefines(self, other):
+        return super(Definition, self).redefines(other) and self.name != "_"
+
 
 class UnhandledKeyType(object):
     """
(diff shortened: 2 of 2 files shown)


sigmavirus24:
This breaks the tests. Care to fix them?

eight04:
Done. It didn't pass trailing space test and python2 test.

sigmavirus24:
Can you add some tests for this?

eight04:
Done.

BTW, the tests failed on my machine. [Here is the result](https://gist.github.com/eight04/0fd5bc4f95df29b486f95391966279a1).

Windows 7 x64
Python 3.5.1

sigmavirus24:
So that's because windows uses `\r\n` for newlines whereas most other operating systems just use `\n`. I'm not sure what the best way to test that is.

I'm a little uncertain we want to accept this pull request for one reason: `_` is often used for the `i18n` translation function, so people might do

```py
from .i18n import _


print(_("Some text to be translated based on locale."))
```

In that case, we probably do not want to silently ignore `_` being undefined in a function. I just don't know what case we want to prefer here. =/

@bitglue do you have any preferences?

ambv:
I agree with **both** of you! IMO, let's ignore redefinitions of `_` *unless it was imported (including imported `as`).

Rationale: `_` is a common throwaway variable. Doing three consecutive for-loops in Python 2 using the `_` name should not raise warnings. However, you never import a thing under a name if you don't intend to use it somehow. In this case, if anything overwrites our imported name (including by a for-loop or other tuple-unpacking throwaway variable), it should definitely raise a warning.

jayvdb:
If we are going to ignore redefinitions(assignments) of `_`, we should also check that `_` is not loaded.

The complication is that in doctest, `_` is magically assigned and is loaded in subsequent lines of code.

sigmavirus24:
@jayvdb I think that's what @ambv and I are talking about, no?

> The complication is that in doctest, _ is magically assigned and is loaded in subsequent lines of code.

Could you provide a link or example? I don't understand what you're talking about here.

jayvdb:
```python
>>> 1
1
>>> _ + 1
2
```

eight04:
What does the building error mean?

asmeurer:
I think it's still fine to use `for _ in range(10)` or some similar in a doctest. Even though it gets assigned automatically it's still a throwaway variable of sorts.

sigmavirus24:
> What does the building error mean?

@eight04 that's not your fault. IIRC @jayvdb added the AppVeyor dependency on PyPy3, so we can let them fix it. (For posterity, it seems setuptools no longer supports Py 3.2-like runtimes.)

pyflakes/test/test_other.py
```diff
@@ -161,6 +161,15 @@ def _(): pass
...
 
+    def test_redefinedUnderscoreImportation(self):
+        """
+        Test that shadowing an underscore importation raises a warning.
+        """
+        self.flakes('''
+        from .i18n import _
+        def _(): pass
+        ''', m.RedefinedWhileUnused)
+
```
sigmavirus24: Can you also add a test that doing

```py
from . import i18n

def i18n(): pass
```

Raises a `RedefinedWhileUnused` warning?

eight04: Isn't it already tested here?
https://github.com/eight04/pyflakes/blob/a11599a8bd25bc9c9ab680b5849f50af6b279222/pyflakes/test/test_imports.py#L281

jayvdb:
As requested, https://github.com/PyCQA/pyflakes/pull/83 fixes Appveyor builds.

sigmavirus24:
Close-reopen to trigger a new AppVeyor build after @jdufresne fixed the issues.

PR #74, opened 2016-07-13
Add PyPy 5.3 and PyPy3 5.2-alpha1 to Travis

jayvdb:
No changes needed to support these recently released versions.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,9 @@ python:
   - 3.5
   - nightly
   - pypy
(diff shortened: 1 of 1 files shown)


jayvdb:
I suspect these will soon become the more used, replacing `pypy` and `pypy3` (when it is out of alpha).
And hasten the death of Python 3.2 .

sigmavirus24:
Looks fine to me. Thanks @jayvdb !

PR #75, opened 2016-07-21
Fix TestMain tests on Windows

jayvdb:
Currently the API test module has failures for TestMain class,
added in f0084592, on Windows as SysStreamCapturing is in
universal newlines mode while its super class IntegrationTests
is using a native console stream with newline=os.linesep.

Add Appveyor CI script as .appveyor.yml,
which can be selected in the Appveyor settings.

### DIFF
--- a/pyflakes/test/test_api.py
+++ b/pyflakes/test/test_api.py
@@ -55,10 +55,30 @@ class Node(object):
 
 class SysStreamCapturing(object):
 
-    """Replaces sys.stdin, sys.stdout and sys.stderr with StringIO objects."""
+    """
+    Context manager capturing sys.stdin, sys.stdout and sys.stderr.
+
+    The file handles are replaced with a StringIO object.
+    On environments that support it, the StringIO object uses newlines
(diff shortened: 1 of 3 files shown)


pyflakes/test/test_api.py
```diff
@@ -55,26 +55,50 @@ def __init__(self, lineno, col_offset=0):
 
 class SysStreamCapturing(object):
 
-    """Replaces sys.stdin, sys.stdout and sys.stderr with StringIO objects."""
+    """
```
sigmavirus24: Any reason why the first line of the docstring isn't on this line too? PEP 257 recommends it be on one line (not that this project explicitly follows that).

jayvdb: My understanding is that PEP 257 doesn't recommend either approach for the summary line (same as opening line, or next line).
This code base very obviously uses the latter convention, which is why I have done it.

sigmavirus24: Fair. I checked the example in the document and skimmed over the portion where it said it can be written like this or on the same line.

PR #76, opened 2016-07-21
Fix PyPy2 Windows IntegrationTests

jayvdb:
IntegrationTests.test_errors is failing on Windows under PyPy2 as its stderr emits `\r\r\n` as the line separator.

Add AppVeyor testing for three PyPy releases.

Also add a test for the other more complex stderr message emitted by the Reporter.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,8 +1,46 @@
 # To activate, change the Appveyor settings to use `.appveyor.yml`.
 install:
-  - python -m pip install tox
+  - python -m pip install --upgrade virtualenv pip setuptools tox
+
+  # Fetch the three main PyPy releases
+  - ps: (New-Object Net.WebClient).DownloadFile('https://bitbucket.org/pypy/pypy/downloads/pypy-2.6.1-win32.zip', "$env:appveyor_build_folder\pypy-2.6.1-win32.zip")
+  - ps: 7z x pypy-2.6.1-win32.zip | Out-Null
+  - move pypy-2.6.1-win32 C:\
+
+  - ps: (New-Object Net.WebClient).DownloadFile('https://bitbucket.org/pypy/pypy/downloads/pypy2-v5.3.1-win32.zip', "$env:appveyor_build_folder\pypy2-v5.3.1-win32.zip")
+  - ps: 7z x pypy2-v5.3.1-win32.zip | Out-Null
+  - move pypy2-v5.3.1-win32 C:\
+
+  - ps: (New-Object Net.WebClient).DownloadFile('https://bitbucket.org/pypy/pypy/downloads/pypy3-2.4.0-win32.zip', "$env:appveyor_build_folder\pypy3-2.4.0-win32.zip")
+  - ps: 7z x pypy3-2.4.0-win32.zip | Out-Null
+  - move pypy3-2.4.0-win32 C:\
+
+  # pypy3 installer provides 'pypy.exe', not pypy3.exe.
+  - copy C:\pypy3-2.4.0-win32\pypy.exe C:\pypy3-2.4.0-win32\pypy3.exe
+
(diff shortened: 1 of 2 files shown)


.appveyor.yml
```diff
@@ -0,0 +1,46 @@
+# To activate, change the Appveyor settings to use `.appveyor.yml`.
+install:
+  - python -m pip install --upgrade virtualenv pip setuptools tox
+
+  # Fetch the three main PyPy releases
+  - ps: (New-Object Net.WebClient).DownloadFile('https://bitbucket.org/pypy/pypy/downloads/pypy-2.6.1-win32.zip', "$env:appveyor_build_folder\pypy-2.6.1-win32.zip")
+  - ps: 7z x pypy-2.6.1-win32.zip | Out-Null
```
sigmavirus24: 2.6 is ... well very old. Why are we supporting that on Windows? That seems wasteful to say the least.

pyflakes/api.py
```diff
@@ -103,7 +103,8 @@ def checkPath(filename, reporter=None):
         return 1
     except IOError:
         msg = sys.exc_info()[1]
-        reporter.unexpectedError(filename, msg.args[1])
+        msg = msg.args[1]
```
sigmavirus24: What's the value in this assignment?

sigmavirus24:
This has merge conflicts. Mind rebasing before I review much more?

pyflakes/api.py
jayvdb: bah. silly inclusion. sorry about that.

.appveyor.yml
jayvdb: Despite its very low number, 2.6.1 is not old.  It was released only a year ago.  That series is/was the stable release.  PyPy 2.6.1 is `sys.version_info` 2.7.10, which is the same as the latest PyPy2 5.x release, and it gets those users past the Python SSL problems.  PyPy 5.x appears to be gaining traction as a replacement for PyPy 2.x.

Also, it costs us almost nothing?  The time to add it to the image is negligible.  There are no Windows specific bugs related to PyPy 2.x series; the only Windows specific bug anywhere in pyflakes is https://bitbucket.org/pypy/pypy/issues/2350 , which also occurs on the latest PyPy2 5.x release, and it is only a problem with our test framework -- not the program end-users are faced with.

As we test against Travis `python: pypy`, which is PyPy2.5 (`sys.version_info` 2.7.8), IMO it makes sense to bring our Windows testing to roughly the same level.

Without evidence otherwise, Windows users have been able to use pyflakes on PyPy2.x , and by adding this to the test matrix we ensure that if/when we do break that, we do it intentionally.

I expect we'd break it when Travis stops providing a 2.5/2.6 by default, or the existing pypy 2.5/2.6 workarounds start to really hurt. (afaik, the only painful aspect is the different column offsets in their error messages, which we should raise a bug in pypy so that problem can whittle away over time.)
I expect PyPy3 2.4 will be the first to be killed.  While it has been added to this test matrix, I would be only too happy to remove it entirely just as soon as they provide a -win release of PyPy3 5.x, as it is Python 3.2, and anybody using that in production a second longer than they need to has bigger problems than their syntax checker no longer working.

pyflakes/test/test_api.py
```diff
@@ -661,6 +668,9 @@ def runPyflakes(self, paths, stdin=None):
         if sys.version_info >= (3,):
             stdout = stdout.decode('utf-8')
             stderr = stderr.decode('utf-8')
+        # Workaround https://bitbucket.org/pypy/pypy/issues/2350
```
jayvdb: looks like there will be a fix in the next release
https://bitbucket.org/pypy/pypy/commits/85ea4f3912ce

sigmavirus24:
I missed that you updated this. Thanks @jayvdb!

PR #77, opened 2016-07-27
Fix typos

jwilk:
Found using [mwic](http://jwilk.net/software/mwic) and [anorack](https://github.com/jwilk/anorack).

### DIFF
--- a/pyflakes/checker.py
+++ b/pyflakes/checker.py
@@ -304,7 +304,7 @@ class ImportationFrom(Importation):
 
 
 class StarImportation(Importation):
-    """A binding created by an 'from x import *' statement."""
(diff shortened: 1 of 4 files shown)


pyflakes/checker.py
```diff
@@ -799,7 +799,7 @@ def on_conditional_branch():
             return
 
         if on_conditional_branch():
-            # We can not predict if this conditional branch is going to
+            # We cannot predict if this conditional branch is going to
```
sigmavirus24: This one is a bit dubious. This can be written either way.

The-Compiler: Seems like "cannot" is [much more common](http://www.oxforddictionaries.com/words/cannot-or-can-not) though.

PR #80, opened 2016-09-08
Fix format string checking

jgarst:
format strings (PEP 498) defines a new AST node, JOINEDSTR which
contains a list of string expressions.

### DIFF
--- a/pyflakes/checker.py
+++ b/pyflakes/checker.py
@@ -937,7 +937,7 @@ class Checker(object):
         MATMULT = ignore
 
     # additional node types
-    COMPREHENSION = KEYWORD = FORMATTEDVALUE = handleChildren
+    COMPREHENSION = KEYWORD = FORMATTEDVALUE = JOINEDSTR = handleChildren
 
     def DICT(self, node):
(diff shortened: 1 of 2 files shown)


jgarst:
Hi - it looks like format strings in 3.6 now define two new AST nodes, FORMATTEDVALUE and JOINEDSTR, which causes pylint to crash in getNodeHandler with 
```AttributeError: 'Checker' object has no attribute 'JOINEDSTR'```

Having pylint pass and handle JOINEDSTR children is something of a guess - my hope is that if there are more conditions that can be checked, you will have a better idea of what those are than I do.

Let me know if you need anything else.

jayvdb:
Thanks @jgarst .

fgimian:
Dear folks, now that Python 3.6.0 is officially released, this particular pull request is more important to ensure correct support of f-strings.  However, the latest official version doesn't include the fix.

Would you kindly consider a new release in the near future for those of us who have jumped on the Python 3.6 bandwagon a little early? 😄 

f-strings are so great, it's hard to go back to using `.format()` 😄 

Thanks so much!
Fotis

PR #83, opened 2016-12-14
Appveyor: Use setuptools 29

jayvdb:
setuptools 30 emits an error when run on Python 3.2 based versions.
Install setuptools 29.0.1 to bootstrap the win32 environment
testing pypy-2.6.1 & pypy3-2.4.0.

PR #84, opened 2016-12-22
Enable support for PEP 526 annotated assignments

ambv:
Without this change, code with annotated assignments is crashing pyflakes with an attribute error (`object has no attribute 'ANNASSIGN'`).

**Test plan**: new tests introduced conforming to behavior described in the PEP.

### DIFF
--- a/pyflakes/checker.py
+++ b/pyflakes/checker.py
@@ -1335,3 +1335,21 @@ class Checker(object):
                 del self.scope[node.name]
             except KeyError:
                 pass
+
+    def ANNASSIGN(self, node):
+        """
+        Annotated assignments don't have annotations evaluated on function
+        scope, hence the custom implementation.
+
+        See: PEP 526.
+        """
+        if node.value:
+            # Only bind the *targets* if the assignment has a value.
+            # Otherwise it's not really ast.Store and shouldn't silence
+            # UndefinedLocal warnings.
+            self.handleNode(node.target, node)
+        if not isinstance(self.scope, FunctionScope):
+            self.handleNode(node.annotation, node)
+        if node.value:
+            # If the assignment has value, handle the *value* now.
+            self.handleNode(node.value, node)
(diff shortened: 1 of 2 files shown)


ambv:
The AppVeyor failure is unrelated (`RuntimeError: Python 3.3 or later is required`, covered in #83).

pyflakes/checker.py
```diff
@@ -1335,3 +1335,19 @@ def EXCEPTHANDLER(self, node):
...
+
+        See: PEP 526.
+        """
+        if node.value:
+            # If the node has no value, it's not really ast.Store and shouldn't
+            # silence UndefinedLocal warnings.
+            self.handleNode(node.target, node)
+        if not isinstance(self.scope, FunctionScope):
+            self.handleNode(node.annotation, node)
+        if node.value:
```
sigmavirus24: Can we add a comment here explaining that this is intentional? I had to read the body a few times (pre-coffee) to recognize that above you're passing `node.target` and here `node.value`.

ambv: FINE.

Agreed, it's little code but has an alarming density of overly generic terms (you're saying I should HANDLE the NODE which is the parent node's TARGET? Alrighty, and now let's HANDLE the NODE which is the parent node's VALUE!) 🤣

ambv:
Ping?

sigmavirus24:
Sorry about that, @ambv. Lost track of this. This looks good to me. Thanks!

ambv:
@bitglue It would be great to get a release with this before @sigmavirus24 releases the next flake8. Let me know if I can help with this somehow. Currently the versions on PyPI are unable to lint type annotated Python 3.6 programs.

bitglue:
I'll try to do that this evening. Feel free to keep bugging me if I forget.

bitglue:
Released 1.5.0 just now.

porterjamesj:
Thank you all for being so on top of these! I really appreciate, it's nice to be able to get my linting working again by just `pip install -U pyflakes`.

ambv:
cc @sigmavirus24 Let's make sure the next flake8 pins pyflakes 1.5.0 ✨

ambv:
Thank you @bitglue! 🍰

PR #85, opened 2016-12-27
Increment version to 1.3.1

nvllsvm:
Since Python 3.6 stable is out, pyflakes should be incremented to release support for PEP 498.


sigmavirus24:
We'll increment this ourselves when we decide to release a new version of PyFlakes. Thanks!

mmedal:
Hey @sigmavirus24, do you have a eta on when you might release 1.3.1? We're on python 3.6, but our automated flake8 checks are throwing exceptions to the tune of the fix in #80.

bitglue:
I released 1.4.0 just now.

PR #86, opened 2017-01-14
Don't test against file perms when running as root

cfs-pure:
test_permissionDenied tests file modes by creating a temporary file, then sets the mode to 0000 and attempts to process it and see if and permission denied error is generated.  This never happens when the unit tests are run as root since it can open files regardless of file permissions.

### DIFF
--- a/pyflakes/test/test_api.py
+++ b/pyflakes/test/test_api.py
@@ -513,6 +513,7 @@ foo = '\\xyz'
             sourcePath, [decoding_error])
 
     @skipIf(sys.platform == 'win32', 'unsupported on Windows')
+    @skipIf(os.getuid() == 0, 'root can do anything')
     def test_permissionDenied(self):
         """
         If the source file is not readable, this is reported on standard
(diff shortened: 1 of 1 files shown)


sigmavirus24:
Closing and reopening to generate a new merged commit for AppVeyor to test.

pyflakes/test/test_api.py
```diff
@@ -513,6 +513,7 @@ def test_invalidEscape(self):
             sourcePath, [decoding_error])
 
     @skipIf(sys.platform == 'win32', 'unsupported on Windows')
+    @skipIf(os.getuid() == 0, 'root can do anything')
```
sigmavirus24: You need to move this inside the function then, e.g.,

```py
@skipIf(sys.platform == 'win32', 'unsupported on Windows')
def test_permissionDenied(self):
    """ ... """
    if os.getuid() == 0:
        self.skipTest('root can do anything')
    # ...
```

At that point, we'll be certain that we're not on Windows, whereas the module scope can't be certain.

sigmavirus24:
@cfs-pure I pushed updates to your branch. We're just waiting for AppVeyor to recover from their temporary outage.

PR #87, opened 2017-01-14
Pip install setuptools

jaraco:
These changes work around pypa/pip#2669 enabling pip to be installed into pypy and avoid using a manual install of setuptools, deprecated in pypa/setuptools#581.


sigmavirus24:
Edited, squashed, and merged in f13b1e4c6473f0e91135765452956055e5091fb7. Thanks @jaraco!

PR #88, opened 2017-01-15
Process function scope variable annotations

jstasiak:
Even though variable annotations in function scope aren't evaluated at
runtime it's still useful for static analysis tools to process them and
catch some issues (and not report some things that aren't issues).

Let's take the following code:

```python
     from typing import Any

     def fun():
          a: Any
```

Previously pyflakes would report Any to be unused:

```
     test.py:1: 'typing.Any' imported but unused
```

With this patch it's no longer the case.

### DIFF
--- a/pyflakes/checker.py
+++ b/pyflakes/checker.py
@@ -1337,19 +1337,12 @@ class Checker(object):
                 pass
 
     def ANNASSIGN(self, node):
-        """
-        Annotated assignments don't have annotations evaluated on function
-        scope, hence the custom implementation.
-
-        See: PEP 526.
-        """
         if node.value:
             # Only bind the *targets* if the assignment has a value.
             # Otherwise it's not really ast.Store and shouldn't silence
             # UndefinedLocal warnings.
             self.handleNode(node.target, node)
-        if not isinstance(self.scope, FunctionScope):
-            self.handleNode(node.annotation, node)
+        self.handleNode(node.annotation, node)
         if node.value:
             # If the assignment has value, handle the *value* now.
             self.handleNode(node.value, node)
--- a/pyflakes/test/test_other.py
+++ b/pyflakes/test/test_other.py
@@ -1839,12 +1839,17 @@ class TestAsyncStatements(TestCase):
             name: str = 'Bob'
             age: int = 18
             foo: not_a_real_type = None
-        ''', m.UnusedVariable, m.UnusedVariable, m.UnusedVariable)
+        ''', m.UnusedVariable, m.UnusedVariable, m.UnusedVariable, m.UndefinedName)
         self.flakes('''
         def f():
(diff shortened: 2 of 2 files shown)


jstasiak:
I'm not sure about those CI tasks being stuck for hours now but as far as I know the CI suite passes with this change.

sigmavirus24:
@jstasiak AppVeyor is suffering from an outage at the moment. That said, the tests still need to pass there for this to be accepted.

jstasiak:
All the tests passed, let me know if there are any modifications I need to make here.

ambv:
Instead of commenting out why ANNASSIGN is a separate method, you could have deleted it entirely.

sigmavirus24:
@ambv Care to send a PR to that effect?

ambv:
Actually, nevermind, we still need it as the first check is still custom: `if node.value: handle node.target`.

asmeurer:
Function scope annotations are ignored completely (so long as they're valid Python). With this PR, you'd get

```py
def test():
    a: fdsf = 1
    return a
print(test())
```

```
test.py:2: undefined name 'fdsf'
```

even though if you run the program, it prints `1`.

It probably makes sense to allow names used in function annotations to be imported without unused warnings, as explicitly importing `Any` from `typing` serves as documentation. But should it fail in the above case? I really don't know what the best practice is, but in a technical sense, this *is* a false positive.

ambv:
Yeah, this is why originally I put the special case for function-scope annotations. However, the unused import warning is clearly disruptive and I changed my mind. What @jstasiak did here is right.

Moreover, the most likely cause of an invalid function-scope annotation will be a typo. The user *wants* to see a warning about it.

The reason that function-scope annotations aren't evaluated is purely efficiency. They would need to be evaluated on every call which is prohibitively costly. The idea to instead "evaluate them on function creation" was rejected by Guido because their presence inside a function clearly suggests they are in function scope. They can also successfully used function-local imports.

Summing up, @asmeurer, I wouldn't worry about the "technically" false positive here. We are doing the right thing now.

dahlia:
When would this fix be released to PyPI?

ambv:
It is part of 1.5.0.

Daenyth:
I'm using 1.5.0 and still experiencing this error.

```
$ pip freeze | grep pyflakes
pyflakes==1.5.0

$ cat foo.py
from typing import Any
def foo(x):
    y: Any = x
    return y

$ python -m pyflakes foo.py
foo.py:1: 'typing.Any' imported but unused
```

myint:
1.5.0 was from 9 January 2017. This pull request was merged in 22 January 2017.

Daenyth:
Ok, I misunderstood what @ambv said to mean that the fix was in 1.5.0

PR #89, opened 2017-02-26
Increment version to 1.5.0

ambv:
Not sure why this is not the case yet? :)


ambv:
The failures on python-nightly are not related.

sigmavirus24:
I suspect @bitglue has this locally on his machine but did not push it.

bitglue:
Yep, thanks.

PR #90, opened 2017-05-12
Allow failure in "nightly" Python until we fix it

myint:


myint:
It looks like the CPython 3.7 failure is related to a change in the AST. The docstring is no longer in `node.body`.

https://github.com/python/cpython/pull/46

We can mostly replace our `getDocstring(node.body[0])` call with `node.docstring`. But I'm not sure how to get the line number of the docstring anymore. `node.lineno` is close, but not quite the line number of the docstring.

https://github.com/PyCQA/pyflakes/blob/074c21da3174bc88b2ec02fea767a436e2db643d/pyflakes/checker.py#L873

```diff
diff --git a/pyflakes/checker.py b/pyflakes/checker.py
index 382574e..2c19533 100644
--- a/pyflakes/checker.py
+++ b/pyflakes/checker.py
@@ -870,7 +870,11 @@ class Checker(object):

     def handleDoctests(self, node):
         try:
-            (docstring, node_lineno) = self.getDocstring(node.body[0])
+            if getattr(node, 'docstring'):
+                docstring = node.docstring
+                node_lineno = ???  # TODO
+            else:
+                (docstring, node_lineno) = self.getDocstring(node.body[0])
             examples = docstring and self._getDoctestExamples(docstring)
         except (ValueError, IndexError):
             # e.g. line 6 of the docstring for <string> has inconsistent
```

myint:
@methane After the change in python/cpython#46, is there still a simple way to get the line number of the docstring? It used to be something like:

```python
node.body[0].value.lineno
```

Thanks

PR #272, opened 2017-05-28
Use new URL for issues

myint:
- [x] Also, could someone with admin access update the banner? I'm not sure what it should point to now. I often point it to PyPI.
- [x] I also noticed that `pyflakes` is one of the only repository in `PyCQA` that ends its description with a period. It might be good to remove it for uniformity.
screen shot 2017-05-28 at 07 17 26
- [ ] We probably want to point the description in [the old Launchpad site](https://launchpad.net/pyflakes) to GitHub. I don't have access to do this.
screen shot 2017-05-28 at 07 27 04

Thanks!

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -61,7 +61,7 @@ Share your feedback and ideas: `subscribe to the mailing-list
 Contributing
 ------------
 
-Issues are tracked on `Launchpad <https://bugs.launchpad.net/pyflakes>`_.
+Issues are tracked on `GitHub <https://github.com/PyCQA/pyflakes/issues>`_.
 
 Patches may be submitted via a `GitHub pull request`_ or via the mailing list
 if you prefer. If you are comfortable doing so, please `rebase your changes`_


sigmavirus24:
So I updated the GitHub banner but I don't have permissions to update the Launchpad tracker.

sigmavirus24:
Also, I don't think @bitglue can update Launchpad either. We'd need @florentx but they've not been around in quite some time.

bitglue:
I was able to register a bug tracker in launchpad:

https://bugs.launchpad.net/bugs/bugtrackers/pyflakes

But I can't figure out how to associate the pyflakes project with it. There's the configure bugs page which seems to offer the option:

https://bugs.launchpad.net/pyflakes/+configure-bugtracker

Yet, the bug tracker I registered is not found there. Maybe I need to wait for it to take effect or something?

bitglue:
Yep, that was it. Just had to complain about it here first.

2017-06-01-071946_1007x533_scrot

bitglue:
So now that the issues are imported on GitHub, what happens to the launchpad issues? Should they be closed with a link to the associated GitHub issue at least so people who opened them can follow the new issue if they want?

Or does Launchpad have some specific integration with GitHub?

sigmavirus24:
There's no integration with GitHub. At this point, I'd suggest closing them all. `launchpadlib` could automate that

PR #275, opened 2017-05-31
Find Python files without extensions

### LINKED ISSUE #149: Please also check python scripts that are not named *.py
*Original report by [morph-debian](https://launchpad.net/~morph-debian) on [Launchpad](https://bugs.launchpad.net/bugs/970465):*

Hello,
as reported on Debian at http://bugs.debian.org/cgi-bin/bugreport.cgi?bug=653890:

"""
It would be great if pyflakes could detect files that are python scripts
but are not named *.py. It would have to read the first two bytes of
each file to check for scripts and then read the rest of the line to
determine if it is a python script. It would need to process stuff like:

#!/usr/bin/python
#!/usr/local/bin/python
#!/usr/bin/env python
#!/usr/bin/python2.6
#!/usr/bin/python3.2
"""

Regards,
Sandro

myint:
Previously, only files ending in `.py` would be detected as Python files while recursing a directory. This   pull request allows for detecting Python files by also looking for shebangs. I'm using the same technique that I've used in `autopep8` for years.

I timed this against the ~7000 files in `/opt/local/Library/Frameworks/Python.framework/Versions/3.6/`.

Three timings before this change:

```
0m23.091s
0m22.792s
0m22.745s
```

Three timings after this change:

```
0m23.046s
0m23.111s
0m22.860s
```

This closes #149.

### DIFF
--- a/pyflakes/api.py
+++ b/pyflakes/api.py
@@ -5,12 +5,16 @@ from __future__ import with_statement
 
 import sys
 import os
+import re
 import _ast
 
 from pyflakes import checker, __version__
 from pyflakes import reporter as modReporter
 
 __all__ = ['check', 'checkPath', 'checkRecursive', 'iterSourceCode', 'main']
+
+
+PYTHON_SHEBANG_REGEX = re.compile(br'^#!.*\bpython[23]?\b\s*$')
 
 
 def check(codeString, filename, reporter=None):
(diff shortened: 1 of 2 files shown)


pyflakes/api.py
```diff
@@ -108,6 +112,25 @@ def checkPath(filename, reporter=None):
...
+
+    try:
+        with open(filename, 'rb') as f:
+            text = f.read(max_bytes)
+            if not text:
+                return False
+    except IOError:
+        return False
+
+    first_line = text.splitlines()[0]
```
bitglue: Does this work on an empty file?

```
>>> ''.splitlines()[0]
IndexError: list index out of range
>>> 
```

myint: Empty files should be caught on line 125.

bitglue: Oh. Derp.

PR #277, opened 2017-06-03
Detect `pythonw` in shebang

myint:
This addresses https://github.com/PyCQA/pyflakes/issues/149#issuecomment-305881232 by @asmeurer.

https://docs.python.org/3/using/mac.html?highlight=pythonw#running-scripts-with-a-gui


asmeurer:
Awesome, thanks.

asmeurer:
I tested it and it works

sigmavirus24:
Thanks for testing @asmeurer!

PR #293, opened 2017-09-12
Add check for unused exception binding in except: block

jdufresne:


jdufresne:
The AppVeyor failure looks unrelated to this change.

PR #303, opened 2017-10-11
Update versions

hugovk:
Python 3.7 is supported (#271, #273), so include it in README.

And update the classifiers to reflect the officially supported versions.
* ~I take it Python 2.5 is still officially supported, but not testable on Travis.~
  * ~Let me know if this is no longer the case!~
* ~I've left out 3.0, 3.1 and 3.2 as unsupported, as 3.2 is testable on Travis but isn't.~
  * ~Let me know if that should be adjusted!~

Edit: now https://github.com/PyCQA/pyflakes/pull/322 is merged, this just adds Python 3.7.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -9,7 +9,7 @@ parsing the source file, not importing it, so it is safe to use on
 modules with side effects.  It's also much faster.
 
 It is `available on PyPI <https://pypi.python.org/pypi/pyflakes>`_
-and it supports all active versions of Python: 2.7 and 3.4 to 3.6.
+and it supports all active versions of Python: 2.7 and 3.4 to 3.7.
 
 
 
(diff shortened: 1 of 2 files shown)


hugovk:
The AppVeyor failure is unrelated to this PR.

hugovk:
Oh, and whilst we're here, might it be time to officially drop EOL Pythons?

image

https://en.wikipedia.org/wiki/CPython#Version_history

myint:
This looks good. Though, I'm not sure if PyPI will accept a 3.7 classifier yet. I recall trying to do something like this years ago for a previously unreleased version of Python and getting an error when trying to upload my package. I could be misremembering though.

hugovk:
@myint Good point. At least 3.7 is in the [Big List](https://pypi.python.org/pypi?%3Aaction=list_classifiers).

hugovk:
Rebased. Now https://github.com/PyCQA/pyflakes/pull/322 is merged, this just adds 3.7.

myint:
Thanks!

https://pypi.python.org/pypi?:action=browse&c=640

PR #309, opened 2017-10-21
 Fix failed AppVeyor web requests

### LINKED ISSUE #308: Disable AppVeyor in pull requests
Or maybe fix it. Currently, it is just noise.

jdufresne:
<s>It is marking all contributions as broken which is not helpful for reviewing.</s>

---

<s>Someone will need to sign into AppVeyor to disable it there.</s>

---

Allow we requests to servers that only support TLS 1.2. Previously failed with the error:

```
Exception calling "DownloadFile" with "2" argument(s): "The request was aborted: Could not create SSL/TLS secure channel."
```

Was marking all contributions as broken which is not helpful for reviewing.

Fixes #308

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,5 +1,7 @@
 # To activate, change the Appveyor settings to use `.appveyor.yml`.
 install:
+  - ps: "[Net.ServicePointManager]::SecurityProtocol = [System.Net.SecurityProtocolType]::Tls12"
+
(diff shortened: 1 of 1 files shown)


jdufresne:
Good call. After some debugging, it looks like the failure occurred when `DownloadFile` contacted a web server that only supported TLS 1.2 and no other SSL/TLS version. I have modified the configuration to only support TLS 1.2 now.

PR #313, opened 2017-11-21
Handle string literal annotations

### LINKED ISSUE #290: Typing.TYPE_CHECKING imports raise F401
```
$ pip install  flake8

```

```
if TYPE_CHECKING:
    from my_module.test_class import MyClass

object: 'MyClass' = None
```
Flake 8 gives the following error in this scenario:

`my_module\test_class.py:4:5: F401 'my_module.test_class.MyClass' imported but unused`

My current work around is this:

```
if TYPE_CHECKING:
    from my_module.test_class import MyClass  # noqa

object: 'MyClass' = None
```

Any better solutions? Couldwe somehow automatically ignore all F401 errors within the `if TYPE_CHECKING:` block?

taion:
Fixes #290

### DIFF
--- a/pyflakes/checker.py
+++ b/pyflakes/checker.py
@@ -922,6 +922,39 @@ class Checker(object):
         self.popScope()
         self.scopeStack = saved_stack
 
+    def handleAnnotation(self, annotation, node):
+        if isinstance(annotation, ast.Str):
+            # Defer handling forward annotation.
+            def handleForwardAnnotation():
+                try:
+                    tree = ast.parse(annotation.s)
+                except SyntaxError:
+                    self.report(
+                        messages.ForwardAnnotationSyntaxError,
+                        node,
+                        annotation.s,
+                    )
+                    return
+
+                body = tree.body
+                if len(body) != 1 or not isinstance(body[0], ast.Expr):
+                    self.report(
+                        messages.ForwardAnnotationSyntaxError,
+                        node,
+                        annotation.s,
+                    )
+                    return
+
+                parsed_annotation = tree.body[0].value
+                for descendant in ast.walk(parsed_annotation):
+                    ast.copy_location(descendant, annotation)
+
+                self.handleNode(parsed_annotation, node)
+
+            self.deferFunction(handleForwardAnnotation)
+        else:
+            self.handleNode(annotation, node)
+
     def ignore(self, node):
         pass
 
@@ -1160,9 +1193,11 @@ class Checker(object):
                 if arg in args[:idx]:
                     self.report(messages.DuplicateArgument, node, arg)
 
-        for child in annotations + defaults:
-            if child:
-                self.handleNode(child, node)
+        for annotation in annotations:
+            self.handleAnnotation(annotation, node)
+
+        for default in defaults:
+            self.handleNode(default, node)
 
         def runFunction():
 
@@ -1375,7 +1410,7 @@ class Checker(object):
             # Otherwise it's not really ast.Store and shouldn't silence
             # UndefinedLocal warnings.
             self.handleNode(node.target, node)
-        self.handleNode(node.annotation, node)
+        self.handleAnnotation(node.annotation, node)
         if node.value:
             # If the assignment has value, handle the *value* now.
             self.handleNode(node.value, node)
--- a/pyflakes/messages.py
+++ b/pyflakes/messages.py
@@ -231,3 +231,11 @@ class AssertTuple(Message):
     Assertion test is a tuple, which are always True.
     """
     message = 'assertion is always true, perhaps remove parentheses?'
+
+
+class ForwardAnnotationSyntaxError(Message):
+    message = 'syntax error in forward annotation %r'
+
+    def __init__(self, filename, loc, annotation):
+        Message.__init__(self, filename, loc)
+        self.message_args = (annotation,)
--- a/pyflakes/test/test_other.py
+++ b/pyflakes/test/test_other.py
@@ -1890,3 +1890,77 @@ class TestAsyncStatements(TestCase):
             class C:
                 foo: not_a_real_type = None
         ''', m.UndefinedName)
+        self.flakes('''
+        from foo import Bar
+        bar: Bar
+        ''')
+        self.flakes('''
+        from foo import Bar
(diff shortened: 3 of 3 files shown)


pyflakes/checker.py
```diff
@@ -1160,9 +1166,11 @@ def addArgs(arglist):
                 if arg in args[:idx]:
                     self.report(messages.DuplicateArgument, node, arg)
 
-        for child in annotations + defaults:
-            if child:
```
taion: I think this was redundant. The check was added in https://github.com/PyCQA/pyflakes/commit/34fd8ba800d288a41bde4c77d6be61effe94e177#diff-f0429ba5bb9de6e83abcf687aca4c012R751, but `handleNode` already checks for `node is None`, and I don't see how `child` can be a falsy value other than `None` here.

The previous revision did not have this extra falsiness check on `defaults`: https://github.com/PyCQA/pyflakes/commit/34fd8ba800d288a41bde4c77d6be61effe94e177#diff-f0429ba5bb9de6e83abcf687aca4c012L750.

taion:
Hold on, this breaks on http://mypy.readthedocs.io/en/latest/kinds_of_types.html#class-name-forward-references.

taion:
This is ready to go now.

taion:
Sorry for the repeated updates. This is now fixed.

myint:
Thanks!

I was testing this against the CPython standard library and it seemed to crash when run against one of the test files ([`test_functools.py`](https://github.com/python/cpython/blob/bdb8315c21825487b54852ff0511fb4881ea2181/Lib/test/test_functools.py)). You can reproduce it by doing the above or more specifically:

```
$ wget https://raw.githubusercontent.com/python/cpython/bdb8315c21825487b54852ff0511fb4881ea2181/Lib/test/test_functools.py

$ ./py/bin/pyflakes test_functools.py
  File "./py/bin/pyflakes", line 11, in <module>
    load_entry_point('pyflakes==1.6.0', 'console_scripts', 'pyflakes')()
  File "/Users/myint/tmp/pyflakes/py/lib/python3.6/site-packages/pyflakes/api.py", line 208, in main
    warnings = checkRecursive(args, reporter)
  File "/Users/myint/tmp/pyflakes/py/lib/python3.6/site-packages/pyflakes/api.py", line 165, in checkRecursive
    warnings += checkPath(sourcePath, reporter)
  File "/Users/myint/tmp/pyflakes/py/lib/python3.6/site-packages/pyflakes/api.py", line 112, in checkPath
    return check(codestr, filename, reporter)
  File "/Users/myint/tmp/pyflakes/py/lib/python3.6/site-packages/pyflakes/api.py", line 73, in check
    w = checker.Checker(tree, filename)
  File "/Users/myint/tmp/pyflakes/py/lib/python3.6/site-packages/pyflakes/checker.py", line 498, in __init__
    self.runDeferred(self._deferredFunctions)
  File "/Users/myint/tmp/pyflakes/py/lib/python3.6/site-packages/pyflakes/checker.py", line 535, in runDeferred
    handler()
  File "/Users/myint/tmp/pyflakes/py/lib/python3.6/site-packages/pyflakes/checker.py", line 929, in handleForwardAnnotation
    parsed = ast.parse(annotation.s).body[0].value
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/ast.py", line 35, in parse
    return compile(source, filename, mode, PyCF_ONLY_AST)
  File "<unknown>", line 1
    This is a new annotation
                ^
SyntaxError: invalid syntax
```

taion:
Ah, my bad. Didn't handle annotations that weren't actually annotations. I've addressed it now and added test cases.

I've also added a new reported error here since there weren't any existing errors that covered this, and I don't think there are any reasons to want illegal annotations in normal code.

If you think this new error is reasonable, could you please advise on whether the naming is appropriate, given that this will be user-facing? The alternative would be to refer to this as a "string literal annotation" rather than a "forward annotation".

taion:
Rebased and squashed.

Can you confirm that you're okay with the naming and the semantics around the new error? Something like `a: 'A B'` or `a: 'A; B'` is legal Python, but I think it's vanishing unlikely to actually be intentionally used.

JackLeo:
when will this be released to pypi?

joar:
It would be super-useful to have this bugfix released, it's been several months since this was merged into master.

jcrada:
Hi, this bug was fixed in 2.0.0, but there is a case in which I still get `F401`. See below.

```python
import typing
if typing.TYPE_CHECKING:
    from .defuzzifier import Defuzzifier

self.defuzzifier: typing.Optional['Defuzzifier'] = None
```
`fuzzylite/variable.py:29:5: F401 '.defuzzifier.Defuzzifier' imported but unused`

jcrada:
I am not sure, but `mypy` complains if I write something that is not a forward annotation. For example, 
```python
self.defuzzifier: typing.Optional['Some comment'] = None
# mypy: fuzzylite/variable.py:126: error: syntax error in type comment

self.defuzzifier: typing.Optional['SomeComment'] = None
# fuzzylite/variable.py:126: error: Name 'SomeComment' is not defined
```

ambv:
@taion, in the type checker context, yes.

mgedmin:
I'm thinking if there should be an option to disable this, similarly how you can disable checking of code inside doctests.

(I'm currently working on a codebase that doesn't use mypy, but uses string annotations as documentation, without the corresponding `from typing import List` etc. imports, as those caused older pyflakes to emit warnings.  Now new pyflakes emits warnings for the annotations...)

taion:
@jcrada 

I'd suggest opening a new issue for that behavior if it's still causing you problems, though the default postponed evaluation of annotations from PEP 563 likely means that this shouldn't be a huge issue going forward.

The fix seems like it could be fairly straightforward – we track whether we're in an annotation context (perhaps with an `annotationDepth` attribute analogous to the current `nodeDepth` attribute), then handle all strings as postponed annotations rather than as just strings.

taion:
I was going to look at this at some point, but I haven't been able to find the time, and this isn't actively causing problems for me.

The lack of that behavior nevertheless does seem like a real bug, though.

PR #314, opened 2017-11-23
Fix docstring for TwoStarredExpressions

taion:


myint:
Is this really an error? I figured it was just a parenthetical remark. Especially given the period at the end.

```
Two or more starred expressions in an assignment (some remark here).
```

sigmavirus24:
I agree. If we want to be clearer, we could put it inside double back-ticks for reStructuredText code formatting.

taion:
Oops, my bad. Misread.

PR #318, opened 2017-12-06
Make unused for loop variables to be reported

firebirdberlin:
Unused for loop variables are curently not reported by pyflakes. There may be valid reasons why a loop variable is not used.  Nevertheless there should be a warning unless the developer names the variable '_' or 'unused' which is a widely used convention for this case, 

My commit passes all tests and adds a few ones for the new behaviour.

### DIFF
--- a/pyflakes/checker.py
+++ b/pyflakes/checker.py
@@ -356,6 +356,12 @@ class Assignment(Binding):
     """
 
 
+class FOR_LOOP_VARIABLE(Binding):
+    """
+    Represents binding a name as a loop variable
+    """
+
+
 class FunctionDefinition(Definition):
     pass
 
@@ -412,6 +418,7 @@ class FunctionScope(Scope):
     usesLocals = False
     alwaysUsed = set(['__tracebackhide__',
                       '__traceback_info__', '__traceback_supplement__'])
+    namesToIgnore = ('_', '_unused')
 
     def __init__(self):
         super(FunctionScope, self).__init__()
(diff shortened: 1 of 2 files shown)


bitglue:
Pyflakes doesn't have warnings, only errors.

https://github.com/PyCQA/pyflakes#design-principles

I imagine adding this behavior would raise a very large number of errors for perfectly sound code just because it doesn't conform to a particular style. Would you disagree?

firebirdberlin:
Desclaring an unused variable never sounds perfect to me. If it is done willingly one should reflect this using an appripriate name as _. 

pyflakes recently added an error for unused assigments in exceptions as in 

```python
    try:
       ...
    except (TypeError, KeyError) as e:
        return None 
```
This is the same situation, but it raises an error. It is just consequent to add my proposed feature.

asmeurer:
The difference is that python requires setting a variable to do a for loop. The names _ and unused aren't universal (I've never seen them before in this context). I personally try to always use descriptive names like `for name in names` even if the variable isn't used (and if there isn't a descriptive name I usually use `i`). Hence I see this as a stylistic issue and thus out of scope for pyflakes.

bitglue:
Indeed this seems like a stylistic issue. Some people use `_` for gettext and would even be harmed by this check.

You might find pylint more to your liking -- It's much more liberal about the kinds of things it will check.

firebirdberlin:
Sorry, but it's not cosmetics. It a potential source of error. Would it be an option to add a command line argument for these things (e.g. --strict ) ?

bitglue:
Sorry, but I don't think that fits with the project's objectives. Pyflakes currently has no command-line arguments. It doesn't need them because it only complains about things which are _definitely_ problems. This is [by design](https://github.com/PyCQA/pyflakes#design-principles). Being able to run pyflakes on a preexisting code base while getting only actionable errors is a distinguishing feature.

You should really check out pylint. If you want to be alerted about things that are _potentially_ errors, I think you'll find it better for your needs. It will warn about this, and _many_ more potential sources of errors. It also has a lot of options.

Another option for you may be a flake8 plugin, like [flake8-bugbear](https://github.com/PyCQA/flake8-bugbear). It looks like it already implements this check as B007.

PR #322, opened 2018-01-05
Drop support for EOL Python <2.7 and 3.2-3.3

### LINKED ISSUE #319: Drop support for EOL Python versions?
I'd like to suggest pyflakes only supports non-EOL Python versions.

Python 2.5, 2.6, 3.2 and 3.3 are EOL and no longer receive security updates.

image

https://github.com/PyCQA/pyflakes#pyflakes says:

> It is available on PyPI and it supports all active versions of Python from 2.5 to 3.6.

It can be argued that 2.5-2.6 and 3.0-3.3 are no longer active and no longer need support.

Here's the pip installs for pyflakes from PyPI for the last month (via `pypinfo --percent --pip --markdown pyflakes pyversion`) showing low or non-existent numbers for the EOL versions.

| python_version | percent | download_count |
| -------------- | ------: | -------------: |
| 2.7            |   54.1% |        582,452 |
| 3.6            |   20.5% |        221,023 |
| 3.5            |   17.2% |        185,122 |
| 3.4            |    6.8% |         72,930 |
| 3.3            |    0.6% |          6,432 |
| 2.6            |    0.4% |          4,789 |
| 3.7            |    0.4% |          4,226 |
| 3.2            |    0.0% |            414 |
| None           |    0.0% |              4 |

Some other PyCQA projects are planning on ending all Python 2 support around mid-2018 when Python 3.7 is due out (https://github.com/PyCQA/meta/issues/19).

hugovk:
Fixes #319.

This builds on #303, drops EOL versions and does some cleanup. It also adds `python_requires` to setup.py to help pip.


.gitignore
```diff
@@ -1,6 +1,7 @@
 *.egg
 *.egg-info
 *.pyc
+.idea
```
bitglue: please add these to a [global gitignore](https://help.github.com/articles/ignoring-files/#create-a-global-gitignore) rather than the project

bitglue:
Also, we should probably merge this and #303 together and release a new major version since it breaks backwards compatibility.

.gitignore
hugovk: Unfortunately, this belongs in the project gitignore and cannot go in the global gitignore because some projects require some of the files from this directory to be in Git.

See https://intellij-support.jetbrains.com/hc/en-us/articles/206544839 for more info.

hugovk:
Appveyor looks like it needs an updated pip to recognise the new python_requires. I had a go updating it in another branch, but there's a pypy bug preventing that. They've just fixed it, and it'll be in the next pypy release, 5.10.1. 


See https://bitbucket.org/pypy/pypy/issues/2720/ensurepip-on-pypy-c-jit-93579-a4194a67868f and https://stackoverflow.com/q/47999518/724176.

I do not know when that will be out. Options are to wait until it's sorted until merge, or comment out that pypy build until then. (Merge with the failure doesn't really help.)

.gitignore
bitglue: Eh? That you've put it in `.gitignore` explicitly states those files should _not_ be tracked in git. IntelliJ is a detail of your preferred development environment, and not Pyflakes. That's why it should go in your personal .gitignore, not Pyflakes.

I use Vim which puts a `.swp` file next to whatever I'm editing. Some people use editors which make backups with a `~` suffix. But those aren't in here either.

bitglue:
Let's just comment that build with an explanation for now.

.gitignore
hugovk: Exactly, I'm saying none of those files from the .idea dir should be in *this* project's repo.

But some *other* projects do want to share IDE config via some files from the .idea dir. So it's not a global level thing, it's a project-level thing. 

I assume Pyflakes does not want any of that config, hence ignoring the entire dir in this case. 

What sort of problems does it cause to have an extra entry in this gitignore?

hugovk:
Commented out!

.gitignore
sigmavirus24: @hugovk when people insist upon thrusting their personal editor's junk into project gitignore files, it invites lots of unwelcome and unhelpful pull requests to add ever more editor junk files to the gitignore, usually by people who otherwise don't contribute to a project. I'm in agreement with @bitglue that this doesn't belong in this project's `.gitignore`. If projects want to track that information, then it can be added inspite of your global gitignore. What your global gitignore prevents, however, is adding it blindly with `git add .`.

bitglue: You can also follow the instructions for a global gitignore above, but omit the `--global` option. This will change the setting just for the repo, so (just for pyflakes) you can use an alternative .gitignore. I'm not sure if that supersedes or arguments the default .gitignore, but you may want to experiment.

You might also be able to put a fully qualified path in your global .gitignore to ignore `.idea` in pyflakes, but not your other git repositories. Again I'm not sure this works, but something to try.

Prepending a leading `!` to a line in gitignore can be used to unignore a file. I'm not familiar with the exact semantics of this feature as I've not had reason to do use it, but perhaps you could add it to projects for which sharing IDE files is explicitly desired.

In any case, the `.gitignore` here does not currently contain any entries that reflect developer preference: only those due to pyflakes and its associated tooling in this repository. That's unlikely to change.

Regardless, thank you for the contributions. They are very much appreciated. I should be able to get this merged and released soon.

hugovk: Aha, thanks for the tips on unignoring the global ignore, I'll give that a shot! I've removed it from this project's one!

PR #324, opened 2018-01-17
Closes https://github.com/PyCQA/pyflakes/issues/249

seeeturtle:


pyflakes/test/test_undefined_names.py
```diff
@@ -326,7 +326,7 @@ def f2():
             global m
         ''', m.UndefinedName)
 
-    @skip("todo")
+    @skip('todo')
```
sigmavirus24: Why change these to single quotes?

seeeturtle: sorry. my mistake

seeeturtle: done

seeeturtle:
@sigmavirus24 it seems the CI is failing not because of the test.
Can you look at it?

seeeturtle:
ah. sorry, I didn't know that it's error flake8
please ignore the above comment

seeeturtle:
wait... all flake8 error isn't related with my PR.
it should be fixed.

pyflakes/checker.py
```diff
@@ -1292,13 +1317,21 @@ def TUPLE(self, node):
...
     def IMPORT(self, node):
         for alias in node.names:
+            name = alias.name
             if '.' in alias.name and not alias.asname:
-                importation = SubmoduleImportation(alias.name, node)
+                importation = SubmoduleImportation(name, node)
             else:
                 name = alias.asname or alias.name
-                importation = Importation(name, node, alias.name)
+                importation = Importation(name, node, name)
```
adhikasp: Are you sure about this? The 3rd arg of `Importation` originally always pass `alias.name` (the fullname of import) and now it could have value of `alias.asname`. The meaning of code changes.

pyflakes/checker.py
```diff
@@ -430,6 +431,13 @@ def unusedAssignments(self):
                     and isinstance(binding, Assignment)):
                 yield name, binding
 
+    def usedAssignments(self):
```
adhikasp: Why is this declared but never used?

pyflakes/checker.py
```diff
@@ -1331,6 +1364,13 @@ def IMPORTFROM(self, node):
                                               module, alias.name)
             self.addBinding(node, importation)
 
+            if isinstance(self.scope, FunctionScope):
```
myint: This code block is the same as the block [above](https://github.com/PyCQA/pyflakes/pull/324/files#diff-f0429ba5bb9de6e83abcf687aca4c012R1328). It should probably be factored out into a separate function.

pyflakes/checker.py
seeeturtle: sorry. my mistake

seeeturtle:
> Also please add the test from the original issue to make sure the problem is corrected.
@adhikasp, I've already added. check the commit message.

pyflakes/checker.py
seeeturtle: yep

pyflakes/checker.py
seeeturtle: oh, sorry. it's my mistake

adhikasp:
> I've already added. check the commit message.

You mean this one? https://github.com/PyCQA/pyflakes/pull/324/commits/fdf2353d4b53c4c1305ed80e31ea6252845c6f26#diff-44d7abf892e78127bd9438c51e1cfd49R340

But this is different compared to the snippet in the original issue. There should be another test when a function have args with same name as a global variable declared inside it. The function in your test above doesn't receive any argument.

seeeturtle:
Sorry. I thought it was same code.

seeeturtle:
@adhikasp By the way, What error should this test catch? `UndefinedName`?

seeeturtle:
The issue is actually different with the given task at gci...

myint:
Yeah, the GCI task I specified is different to #249. The one I specified matches the test case @seeeturtle created. So I think @adhikasp's comment about the test doesn't apply.

(Actually, #249 would have been easier. I probably should have specified that case for the GCI task.)

seeeturtle:
@myint Can you approve the task too?

seeeturtle:
Hooray! I didn't thought I can do this task well, but I did :)

adhikasp:
Oh I see, my bad then 😄

seeeturtle:
Just waiting for February now :) See you then

seeeturtle:
@myint, I've just found that this PR hasn't merged for a while.
Is there any problem?

myint:
@seeeturtle One of the other PyCQA members needs to approve this before I can merge it. This is the case for several of the [open pull requests](https://github.com/PyCQA/pyflakes/pulls).

seeeturtle:
Ah, I got it. Thanks

2018. 7. 16. 오전 1:16 Steven Myint <notifications@github.com> 작성:

pyflakes/checker.py
```diff
@@ -1292,13 +1319,16 @@ def TUPLE(self, node):
...
+            name = alias.name
             if '.' in alias.name and not alias.asname:
-                importation = SubmoduleImportation(alias.name, node)
+                importation = SubmoduleImportation(name, node)
             else:
                 name = alias.asname or alias.name
                 importation = Importation(name, node, alias.name)
             self.addBinding(node, importation)
 
+            self.store_global_scope(name, alias)
```
jayvdb: This needs to be `self.store_global_scope(importation.name, alias)` , otherwise the following will fail:

```py
        self.flakes('''
            def f(): global foo; import foo.bar
            def g(): foo.is_used()
        ''')
```

You can add that test into `test_assignedToGlobal`.

Importing `foo.bar` is a bit strange -- it imports and binds the module `foo` as that name, and then imports `foo.bar` but there is no additional binding in the scope for `foo.bar`.

The determination of the binding name is done within `SubmoduleImportation`.

Note it would be much better if you did the `self.store_global_scope(...)` inside `addBinding`

pyflakes/checker.py
```diff
@@ -1292,13 +1319,16 @@ def TUPLE(self, node):
 
     def IMPORT(self, node):
         for alias in node.names:
+            name = alias.name
             if '.' in alias.name and not alias.asname:
-                importation = SubmoduleImportation(alias.name, node)
+                importation = SubmoduleImportation(name, node)
```
jayvdb: this change can be reverted.

pyflakes/checker.py
```diff
@@ -1074,17 +1085,33 @@ def NAME(self, node):
...
+                if node.id not in global_scope:
+                    self.report(messages.UndefinedName, node, node.id)
         elif isinstance(node.ctx, (ast.Store, ast.AugStore)):
             self.handleNodeStore(node)
+            if (isinstance(self.scope, FunctionScope) and
+                    node.id in self.scope.global_names):
+                global_scope.setdefault(node.id, Assignment(node.id, node))
         elif isinstance(node.ctx, ast.Del):
             self.handleNodeDelete(node)
+            if (isinstance(self.scope, FunctionScope) and
```
jayvdb: this should go in `handleNodeDelete`, where there is already logic regarding globals

pyflakes/checker.py
```diff
@@ -1074,17 +1085,33 @@ def NAME(self, node):
...
                     and isinstance(node.parent, ast.Call)):
                 # we are doing locals() call in current scope
                 self.scope.usesLocals = True
+            if (isinstance(self.scope, FunctionScope) and
+                    node.id in self.scope.global_names):
+                if node.id not in global_scope:
+                    self.report(messages.UndefinedName, node, node.id)
         elif isinstance(node.ctx, (ast.Store, ast.AugStore)):
             self.handleNodeStore(node)
+            if (isinstance(self.scope, FunctionScope) and
```
jayvdb: this should go in `handleNodeStore`, where there is already logic regarding globals

pyflakes/checker.py
```diff
@@ -1074,17 +1085,33 @@ def NAME(self, node):
...
+        global_scope_index = 1 if self._in_doctest() else 0
+        global_scope = self.scopeStack[global_scope_index]
         # Locate the name in locals / function / globals scopes.
         if isinstance(node.ctx, (ast.Load, ast.AugLoad)):
             self.handleNodeLoad(node)
             if (node.id == 'locals' and isinstance(self.scope, FunctionScope)
                     and isinstance(node.parent, ast.Call)):
                 # we are doing locals() call in current scope
                 self.scope.usesLocals = True
+            if (isinstance(self.scope, FunctionScope) and
```
jayvdb: this should probably go in `handleNodeLoad`

pyflakes/checker.py
seeeturtle: Sorry for responding too late, @jayvdb.
I was reading the `addBinding` and there is something I can't understand.
Where does `scope` comes from? I can't find it...

Anyway, Thanks for reviewing :)

seeeturtle: self-answering...

if break statement is occured, the variable in for loop can be used at outside of it.

retnikt:
Will this ever get merged? It's been two years!

PR #325, opened 2018-01-19
Support `__class__` without `self` in Python 3

### LINKED ISSUE #223: On Python 3, __class__ can be used without self
*Original report by [icordasc](https://launchpad.net/~icordasc) (@sigmavirus24?) on [Launchpad](https://bugs.launchpad.net/bugs/1487725):*

On Python 3, the following code raises a warning:

```python
class Test(object):
    def __init__(self):
        print(__class__.__name__)
        self.x = 1

t = Test()
```

This is actually valid Python 3 code, but PyFlakes doesn't quite understand that:

    test.py:3: undefined name '__class__'

myint:
This fixes #223.

I created this patch previously in:

https://github.com/PyCQA/pyflakes/commit/83ab002d6fab2b717df91e854624c6fdfd9b9213

But I deleted that branch for some reason. I can't recall why I disliked it enough to delete it. But since I can't remember, I'll commit it to a branch again for others to look at.

### DIFF
--- a/pyflakes/checker.py
+++ b/pyflakes/checker.py
@@ -710,10 +710,14 @@ class Checker(object):
 
         # try enclosing function scopes and global scope
         for scope in self.scopeStack[-1::-1]:
-            # only generators used in a class scope can access the names
-            # of the class. this is skipped during the first iteration
-            if in_generators is False and isinstance(scope, ClassScope):
-                continue
+            if isinstance(scope, ClassScope):
+                if not PY2 and name == '__class__':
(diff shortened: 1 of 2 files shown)


myint:
I still don't see anything wrong with this. I'll merge it.

graingert:
@bitglue can we get a release of pyflakes with this change?

It might also be worth adding @sigmavirus24 as maintainer on PyPI?

bitglue:
Released.

PR #328, opened 2018-01-29
ENH: Warn on raise NotImplemented(...).

ssanderson:
This is a common mistake for `raise NotImplementedError`. Since
`NotImplemented` isn't an exception, it's never valid to raise
it (barring strange circumstances where you've aliased `NotImplemented`
in `__builtins__`).

### DIFF
--- a/pyflakes/checker.py
+++ b/pyflakes/checker.py
@@ -965,7 +965,7 @@ class Checker(object):
 
     # "stmt" type nodes
     DELETE = PRINT = FOR = ASYNCFOR = WHILE = IF = WITH = WITHITEM = \
-        ASYNCWITH = ASYNCWITHITEM = RAISE = TRYFINALLY = EXEC = \
+        ASYNCWITH = ASYNCWITHITEM = TRYFINALLY = EXEC = \
         EXPR = ASSIGN = handleChildren
 
     PASS = ignore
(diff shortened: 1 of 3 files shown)


pyflakes/checker.py
```diff
@@ -989,6 +989,22 @@ def ignore(self, node):
         EQ = NOTEQ = LT = LTE = GT = GTE = IS = ISNOT = IN = NOTIN = \
         MATMULT = ignore
 
+    def RAISE(self, node):
+        self.handleChildren(node)
+
+        # Check for `
+        raise_arg = node.exc
+        if not isinstance(raise_arg, ast.Call):
```
bitglue: What about `raise NotImplemented`?

ssanderson:
@bitglue yup, looks like the ast nodes are different in python 2/3.  I'll look into the failures tonight.

pyflakes/checker.py
ssanderson: good point. I can make this look for an `ast.Name` with an id of "NotImplemented" as well.

ssanderson:
@bitglue this took me a bit longer to get back to than I had hoped, but I think this should be good for another look.  The py2 failures were because the argument to `ast.Raise` is `type` in python 2, but it's `exc` in python 3.  I also made this handle bare `raise NotImplemented`.

PR #329, opened 2018-03-13
checker.py: Check for invalid print syntax

MacBox7:
New warning added in messages.py to check for invalid
python syntax

Closes https://github.com/PyCQA/pyflakes/issues/242

### DIFF
--- a/pyflakes/checker.py
+++ b/pyflakes/checker.py
@@ -768,6 +768,13 @@ class Checker(object):
                     # iteration
                     continue
 
+            if (name == 'print' and
+                    isinstance(scope.get(name, None), Builtin)):
+                parent = self.getParent(node)
+                if (isinstance(parent, ast.BinOp) and
+                        isinstance(parent.op, ast.RShift)):
+                    self.report(messages.InvalidPrintSyntax, node)
+
             try:
                 scope[name].used = (self.scope, node)
             except KeyError:
--- a/pyflakes/test/test_other.py
+++ b/pyflakes/test/test_other.py
@@ -2034,3 +2034,53 @@ class TestAsyncStatements(TestCase):
         self.flakes('''
         raise NotImplemented
         ''', m.RaiseNotImplemented)
+
+
+class TestIncompatiblePrintOperator(TestCase):
+    """
+    Tests for warning about invalid use of print function.
+    """
+
+    def test_valid_print(self):
+        self.flakes('''
+        print("Hello")
+        ''')
+
+    def test_invalid_print_when_imported_from_future(self):
+        exc = self.flakes('''
+        from __future__ import print_function
+        import sys
+        print >>sys.stderr, "Hello"
+        ''', m.InvalidPrintSyntax).messages[0]
+
+        self.assertEqual(exc.lineno, 4)
+        self.assertEqual(exc.col, 0)
+
+    def test_print_function_assignment(self):
+        """
+        A valid assignment, tested for catching false positives.
+        """
+        self.flakes('''
+        from __future__ import print_function
+        log = print
+        log("Hello")
+        ''')
+
+    def test_print_in_lambda(self):
+        self.flakes('''
+        from __future__ import print_function
+        a = lambda: print
+        ''')
+
+    def test_print_returned_in_function(self):
+        self.flakes('''
+        from __future__ import print_function
+        def a():
+            return print
+        ''')
+
+    def test_print_as_condition_test(self):
+        self.flakes('''
+        from __future__ import print_function
+        if print: pass
+        ''')
--- a/pyflakes/messages.py
+++ b/pyflakes/messages.py
@@ -250,3 +250,7 @@ class ForwardAnnotationSyntaxError(Message):
 
 class RaiseNotImplemented(Message):
     message = "'raise NotImplemented' should be 'raise NotImplementedError'"
+
+
+class InvalidPrintSyntax(Message):
+    message = 'use of >> is invalid with print function'


myint:
I see a false positive on Python 3:

`foo.py`:
```python
def main():
    log = print
    log('hello')
```

```
$ python -m pyflakes foo.py
foo.py:2: incompatible use of print operator
```

MacBox7:
TypeErrors are detected on execution. What we can do is list out all possible usecase where print won't throw an error and report all others. Could there be a more general approach to do this?

For eg: 
We can whitelist print if its parent is ast.Call or ast.Assign and report all other instance.

pyflakes/checker.py
```diff
@@ -716,6 +716,12 @@ def handleNodeLoad(self, node):
         if not name:
             return
 
+        if name == 'print':
```
sigmavirus24: This needs tests. Further, it looks as though this is running on Python 2 and 3, so someone running pyflakes on Python 2 against Python 2-only code will see a false, positive, right?

MacBox7: No the ast structure for `print` is different. In python2 it is a seperate `Node` however in python3, since it is a function it comes under `Name` node.

MacBox7:
`Hacking` and `flake8_print` also shows false positive. So, either we can flag every occurent where print is not used as a function or we can stick to the approach that I have updated.

MacBox7:
@sigmavirus24 i understand and hence took a different approach which takes into consideration the false positives.

myint:
The latest commit addresses my concern based on my local testing. But it would be good to have test cases to verify this. Thanks!

MacBox7:
@myint, as I said detecting TypeError with static analysis, is a challenge. We can rely on this heuristic approach to solve the problem. Plus, I don't think of any other case where it might fail. I have already added a test case for this. Do you want me to add more?

pyflakes/test/test_other.py
```diff
@@ -2000,3 +2000,16 @@ def test_raise_notimplemented(self):
         self.flakes('''
         raise NotImplemented
         ''', m.RaiseNotImplemented)
+
+
+class TestAdHocConditions(TestCase):
```
sigmavirus24: What does "AdHocConditions" mean here?

Aren't we just testing the cases where print can be used? Can we come up with a better name please?

pyflakes/test/test_other.py
```diff
@@ -2000,3 +2000,16 @@ def test_raise_notimplemented(self):
...
+    def test_invalidPrintFunctions(self):
+            """
+            print keyword is invalid in Python3.x and if print_function
+            is imported from __future__ in Python2.7
+            """
+            self.flakes('''
+            from __future__ import print_function
+            import sys
+            print >>sys.stderr, "Hello"
+            ''', m.InvalidPrintSyntax)
```
sigmavirus24: I'd advocate for more tests here, and matching the indentation of the rest of this file, e.g.,

```py
self.falkes('''
from __future__ import print_function
log = print
log("Hello")
''')

self.flakes('''
print("Hello")
''')
```

MacBox7:
@sigmavirus24 added some more test cases and indented the code as you requested :smile:

myint:
I don't think I mentioned any `TypeError`. But the latest test fulfills want I was thinking of. Thanks.

MacBox7:
The use case that was mentioned in the issue, was a type error when compiled using compiler.

pyflakes/test/test_other.py
```diff
@@ -2000,3 +2000,33 @@ def test_raise_notimplemented(self):
...
+    @skipIf(version_info < (3,), 'syntax error thrown for Python 2.x')
+    def test_printFunctionAssignment(self):
+        """
+        A valid assignment, tested for catching false positives
+        """
+        self.flakes('''
+        from __future__ import print_function
+        log = print
+        log("Hello")
+        ''')
```
jayvdb: EOF

pyflakes/test/test_other.py
```diff
@@ -2000,3 +2000,33 @@ def test_raise_notimplemented(self):
...
         raise NotImplemented
         ''', m.RaiseNotImplemented)
+
+
+class TestIncompatiblePrintOperator(TestCase):
+    """
+    Tests for warning about invalid print operators.
+    """
+
+    @skipIf(version_info < (3,), 'syntax error thrown for Python 2.x')
```
jayvdb: on my py27, this passes with your patch applied

pyflakes/test/test_other.py
```diff
@@ -2000,3 +2000,33 @@ def test_raise_notimplemented(self):
...
+        ''')
+
+    def test_invalidPrintWhenImportedFromFuture(self):
+        self.flakes('''
+        from __future__ import print_function
+        import sys
+        print >>sys.stderr, "Hello"
+        ''', m.InvalidPrintSyntax)
+
+    @skipIf(version_info < (3,), 'syntax error thrown for Python 2.x')
```
jayvdb: again, on py27 this passes.

pyflakes/checker.py
jayvdb: You need to add another bunch of tests without `__future__.print_function` , run only on py27 , to capture their results, and ideally we make those results similar to the errors emitted with print_function enabled.

pyflakes/test/test_other.py
MacBox7: Resolved

pyflakes/checker.py
jayvdb: Actually not needed.  Thx

pyflakes/checker.py
```diff
@@ -716,6 +716,11 @@ def handleNodeLoad(self, node):
...
+            whitelisted_parents = (ast.Call, ast.Assign)
```
jayvdb: by whitelisting only a limited set of parent nodes, it becomes invalid in lots of other scenarios than the original issue was about.

e.g.
```py
def a():
    return print

a = lambda: print

if print: pass
```

I have seen valid code which does the first case above, in code which how a program should output error info.  It isnt hard to imagine the other two occurring in real code;  for example, any code which uses `print = log`.

There may be others which also need to be whitelisted.

It might be better to define the set of cases where `print` is semantically ridiculous, and specifically exclude them.

Either way will require some more tests to show which usage types are valid and which are not valid.

asmeurer:
If the goal is to catch the case specified in the issue, why not check for it exactly? I think print as the left operand of an lshift is the only way a Python 2 `print` can be valid in Python 3. 

Also, this should check if `print` has been redefined and disable the test if it has.

pyflakes/checker.py
```diff
@@ -768,6 +768,13 @@ def handleNodeLoad(self, node):
                     # iteration
                     continue
 
+            if (name == 'print' and
+                    isinstance(scope.get(name, None), Builtin)):
+                parent = self.getParent(node)
+                if (isinstance(parent, ast.BinOp) and
+                        isinstance(parent.op, ast.RShift)):
+                    self.report(messages.InvalidPrintSyntax, node)
```
jayvdb: The reported error node should be the `>>`, not the `print`, as it is the `>>` which is incompatible, and under Python 3 there is no ability to change to using the print operator so that the `>>` can become valid.

jayvdb: Turns out this was impossible to do as the >> node position is not available in the ast.

PR #330, opened 2018-03-14
WIP: Parse type annotations

jdahlin:
Here is a proof of concept of parsing type annotations.

It adds a dependency which is implicit and needs Python 3.x to work, since typed_ast does not work on Python 2.7 (not properly checked).

If you think this is worthwhile I can spend some more time fixing it up so that it checks for the right modules and python versions.

This should fix issue 247.

### DIFF
--- a/pyflakes/checker.py
+++ b/pyflakes/checker.py
@@ -72,6 +72,11 @@ else:
     LOOP_TYPES = (ast.While, ast.For, ast.AsyncFor)
 
 
+def ast_compile(codeString, filename):
+    tree = compile(codeString, filename, 'exec', ast.PyCF_ONLY_AST)
+    return tree
+
+
 class _FieldsOrder(dict):
     """Fix order of AST node fields."""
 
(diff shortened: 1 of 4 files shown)


pyflakes/checker.py
```diff
@@ -873,6 +873,28 @@ def getDocstring(self, node):
...
+                            yield part.strip()
+                    else:
+                        yield token
+            else:
+                yield typeComment
+
+        typeComment = getattr(node, 'type_comment', None)
+        if typeComment is not None:
+            for comment in extractTypeComments(typeComment):
+                typeNode = ast.Name(comment, ast.Load())
```
jdahlin: I realize that this is likely not the best way to create a node, but it worked. Happy for suggestions on how to do this better.

pyflakes/checker.py
```diff
@@ -887,6 +909,7 @@ def handleNode(self, node, parent):
         self.nodeDepth += 1
         node.depth = self.nodeDepth
         node.parent = parent
+        self.handleTypeComments(node)
```
jdahlin: This didn't feel right as well, but worked. Again happy for improvement suggestions.

myint:
I realize this is just for a proof of concept. But is the `typed_ast` dependency only necessary to try to support checking this under Python 2? Would that go away if we wanted to only support type checking under Python 3.

asottile:
via #400

PR #331, opened 2018-03-19
skip Emacs/vim backup files

smontanaro:
isPythonFile should ignore obvious editor backup files. This pull request handles Emacs and vim products.

### DIFF
--- a/pyflakes/api.py
+++ b/pyflakes/api.py
@@ -117,6 +117,10 @@ def isPythonFile(filename):
     if filename.endswith('.py'):
         return True
 
+    # Avoid obvious backup files (Emacs & vim)
+    if filename.endswith("~"):
+        return False
+
     max_bytes = 128
 
     try:
--- a/pyflakes/api.py
+++ b/pyflakes/api.py
@@ -117,7 +117,7 @@ def isPythonFile(filename):
     if filename.endswith('.py'):
         return True
 
-    # Avoid obvious backup files (Emacs & vim)
+    # Avoid obvious Emacs backup files
     if filename.endswith("~"):
         return False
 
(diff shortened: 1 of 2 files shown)


pyflakes/api.py
```diff
@@ -117,6 +117,10 @@ def isPythonFile(filename):
     if filename.endswith('.py'):
         return True
 
+    # Avoid obvious backup files (Emacs & vim)
```
myint: I think this comment needs to be adjusted to only mention Emacs. Vim doesn't do this.

(Vim creates hidden files, but deletes them immediately upon exit. And it is binary data that would not trigger our Python shebang regex. And NeoVim doesn't scatter temporary files in the working directory at all.)

smontanaro: That's fine. I'm not a vim user. I was going by what I saw in this SO thread:

https://stackoverflow.com/questions/607435/why-does-vim-save-files-with-a-extension

That's the first Google hit when you search for "vim backup files". I assume that with the accreditation of both Google and Stack Overflow it must have been authoritative. :-)

smontanaro: I pushed a change to my fork, but don't know how to update the pull request. I don't see an obvious "update pull request" link.

mgedmin: > Vim creates hidden files, but deletes them immediately upon exit. And it is binary data that would not trigger our Python shebang regex. And NeoVim doesn't scatter temporary files in the working directory at all.)

Those are Vim swap files.  Vim creates backup files named `foo.py~` when you do `:w` if you have `:set backup` enabled in your `.vimrc`.

myint: Interesting! I hadn't seen that "feature" before. I usually just get annoyed when seeing those tilde files. 😄

myint:
Thanks!

PR #334, opened 2018-06-08
Don't warn about unused variables which begin with a single underscore.

smontanaro:
This hopefully will address https://github.com/PyCQA/pyflakes/issues/202


pyflakes/checker.py
```diff
@@ -425,7 +425,9 @@ def unusedAssignments(self):
         Return a generator for the assignments which have not been used.
         """
         for name, binding in self.items():
-            if (not binding.used and name not in self.globals
+            if (not binding.used
+                    and (name[0:1] != '_' or name[0:2] == '__')
```
myint: I'm uncertain about the (`name[0:1] != '_' or name[0:2] == '__'`) logic here. One is using `!=` while the other is using a `==`. Can you add additional test case to demonstrate what this is trying to do?

smontanaro:
Sorry, that clause definitely deserved a comment. I originally had this:
```
...
and name[0] != '_'
...
```
but there is another test case which asserts that defining "__all__" inside a local scope without using it is an error. Personally, i wouldn't consider any unused dunder dunder variable to be an error, but I didn't want to break that test case either, so I changed the above to
```
...
and (name[0:1] != '_' or name[0:2] == '__')
...
```
to get the __all__ test case to pass again.

I'm happy to tweak things. Let me know...

pyflakes/checker.py
smontanaro: Sorry added a comment elsewhere. There is another test case which fails without the second half of the or clause. It explicitly tests that __all__ defined in a local scope and not used should be an error. The bit after the 'or' was simply to get that test to pass again.

asmeurer:
#202 is about variables that are just `_`. I don't think skipping variables that just start with `_` was discussed. I don't think it should be done, as that is often used to indicate certain things in Python (e.g. that a variable is throwaway), which don't mean that the variable is unused. Blocking this check for such variables would be an unnecessary regression in my opinion.

pyflakes/checker.py
asmeurer: I don't think this should be done as noted in another comment, but the clean way to write this is to use the `name.startswith` method.

krader1961:
I agree with @asmeurer. A leading underscore is simply a convention that the function, method, or var is private, an implementation detail, not part of the public API of the module or class. It says absolutely nothing about the symbol not being used. Only the magic `_` symbol is special in that regard. And only because it is a widely used convention at companies like Google:

```
def my_func(param_a, param_b):
    _ = param_b  # this param is not used
```

smontanaro:
When all is said and done, I want to give meaningful names even to unused variables. If I call a function which returns a two-element tuple and only need the first for whatever I'm doing, I still want to be reminded at a glance what the meaning of the unused variable is. "_" is not sufficient.

krader1961:
@smontanaro I think you're missing the point. The established idiom that many companies and projects use is to treat assignment to `_` as a way to convey, to humans and lint tools, that you really meant not to use a given variable. The variable that is being unused should have a meaningful name. As for your example I think it is silly but, again, all you have to do is this:

```
(var_a, var_b) = some_func()
_ = var_b
```

We definitely do not want to have the check for being unused suppressed for every var that begins with an underscore.

I vote a strong thumbs down for this PR.

smontanaro:
And I think you're missing my point. I understand that it's possible to accidentally type "_zzyzx" when you really meant to type "zzyzx" or "__zyzzyx", so there will be some cases (rare, I think) where the change I proposed will miss a typo.

A local variable is already effectively "private." There is no need to precede it with an underscore to make it so. (Using a single leading underscore to convey privacy is only useful for otherwise visible attributes of other objects to convey, "Hey, this is my private attribute. Keep your mitts off.")

Ultimately, I think it boils down to the lack of configurability of pyflakes. There are plausible reasons to support both interpretations (sometimes both at different times or at different locations for the same piece of code), but it can only be one way. With pylint, despite the many false positives as it generates, its rich check suppression mechanism lets me have it both ways.

This clearly isn't my code base, so feel free to do what you want. You've had an open ticket for four years which took me a few minutes to fix the way I would have. If all you want to do is support a lone underscore, tweak the clause I added to restrict it to that case and be done with it.

krader1961:
I just noticed this in your change:

```
and (not name.startswith('_')
     or name.startswith('__'))
```

The `or` clause is completely superfluous.  And that should have been obvious to you.

Issue #338, opened 2018-06-16
Handler for ARGUMENTS node missing


MacBox7:
For a code like this:
```
def foo():
    def bar():
        pass
```
while generating pyflakes AST using
```
tree = ast.parse(code)
result = Checker(tree.body[0], filename='test.py')
```
pyflakes result in error:
```
Traceback (most recent call last):
  File "/home/ankit/Documents/GitHub/pyflakes/pyflakes/checker.py", line 1489, in <module>
    result = Checker(tree.body[0], filename='test.py')
  File "/home/ankit/Documents/GitHub/pyflakes/pyflakes/checker.py", line 511, in __init__
    self.handleChildren(tree)
  File "/home/ankit/Documents/GitHub/pyflakes/pyflakes/checker.py", line 861, in handleChildren
    self.handleNode(node, tree)
  File "/home/ankit/Documents/GitHub/pyflakes/pyflakes/checker.py", line 907, in handleNode
    handler = self.getNodeHandler(node.__class__)
  File "/home/ankit/Documents/GitHub/pyflakes/pyflakes/checker.py", line 736, in getNodeHandler
    self._nodeHandlers[node_class] = handler = getattr(self, nodeType)
AttributeError: 'Checker' object has no attribute 'ARGUMENTS'
```

The solution is to either provide a handler or simply ignore it.
Solution is similar like [this](https://github.com/PyCQA/pyflakes/blob/master/pyflakes/checker.py#L971)
```
ARGUMENTS = ignore
```

MacBox7:
Note: Here I am only generating pyflakes AST for a segment of code i.e the inner content of function `foo`.

MacBox7:
As a matter of fact there are quite some nodes defined in [green tree docs](http://greentreesnakes.readthedocs.io/en/latest/nodes.html) whose handler isn't provided. I believe the default mechanism should be to ignore if no handler is present.

myint:
I'm trying to understand the context here. Is this specific to checking a segment of code or can this be reproduced in the general case of checking a full file?

MacBox7:
This is specific to segment of code. I believe there is no harm in in returning pyflakes AST for a segment of code if desired. Plus the work done over [here](https://github.com/PyCQA/pyflakes/blob/master/pyflakes/checker.py#L1186) should be ideally placed in an ARGUMENT handler to improve cohesion.

myint:
[...]

PR #339, opened 2018-06-27
Handle subtrees as input to checker

MacBox7:
Closes https://github.com/PyCQA/pyflakes/issues/338

### DIFF
--- a/pyflakes/checker.py
+++ b/pyflakes/checker.py
@@ -95,9 +95,15 @@ def iter_child_nodes(node, omit=None, _fields_order=_FieldsOrder()):
     """
     Yield all direct child nodes of *node*, that is, all fields that
     are nodes and all items of fields that are lists of nodes.
+
+    :param node:          AST node to be iterated upon
+    :param omit:          String or tuple of strings denoting the
+                          attributes of the node to be omitted from
+                          further parsing
+    :param _fields_order: Order of AST node fields
     """
     for name in _fields_order[node.__class__]:
-        if name == omit:
+        if omit and name in omit:
             continue
         field = getattr(node, name, None)
         if isinstance(field, ast.AST):
@@ -472,6 +478,17 @@ class Checker(object):
         callables which are deferred assignment checks.
     """
 
+    _ast_node_scope = {
+        ast.Module: ModuleScope,
+        ast.ClassDef: ClassScope,
+        ast.FunctionDef: FunctionScope,
+        ast.Lambda: FunctionScope,
+        ast.ListComp: GeneratorScope,
+        ast.SetComp: GeneratorScope,
+        ast.GeneratorExp: GeneratorScope,
+        ast.DictComp: GeneratorScope,
+    }
+
     nodeDepth = 0
     offset = None
     traceTree = False
@@ -493,7 +510,10 @@ class Checker(object):
         if builtins:
             self.builtIns = self.builtIns.union(builtins)
         self.withDoctest = withDoctest
-        self.scopeStack = [ModuleScope()]
+        try:
+            self.scopeStack = [Checker._ast_node_scope[type(tree)]()]
+        except KeyError:
+            raise RuntimeError('No scope implemented for the node %r' % tree)
         self.exceptHandlers = [()]
         self.root = tree
         self.handleChildren(tree)
@@ -643,6 +663,18 @@ class Checker(object):
                 return True
         return False
 
+    def _getAncestor(self, node, ancestor_type):
+        parent = node
+        while True:
+            if parent is self.root:
+                return None
+            parent = self.getParent(parent)
+            if isinstance(parent, ancestor_type):
+                return parent
+
+    def getScopeNode(self, node):
+        return self._getAncestor(node, tuple(Checker._ast_node_scope.keys()))
+
     def differentForks(self, lnode, rnode):
         """True, if lnode and rnode are located on different forks of IF/TRY"""
         ancestor = self.getCommonAncestor(lnode, rnode, self.root)
@@ -790,6 +822,8 @@ class Checker(object):
             binding = Binding(name, node)
         elif name == '__all__' and isinstance(self.scope, ModuleScope):
             binding = ExportBinding(name, node.parent, self.scope)
+        elif isinstance(getattr(node, 'ctx', None), ast.Param):
+            binding = Argument(name, self.getScopeNode(node))
         else:
             binding = Assignment(name, node)
         self.addBinding(node, binding)
@@ -1103,13 +1137,12 @@ class Checker(object):
                     and isinstance(node.parent, ast.Call)):
                 # we are doing locals() call in current scope
                 self.scope.usesLocals = True
-        elif isinstance(node.ctx, (ast.Store, ast.AugStore)):
+        elif isinstance(node.ctx, (ast.Store, ast.AugStore, ast.Param)):
             self.handleNodeStore(node)
         elif isinstance(node.ctx, ast.Del):
             self.handleNodeDelete(node)
         else:
-            # must be a Param context -- this only happens for names in function
-            # arguments, but these aren't dispatched through here
+            # Unknown context
             raise RuntimeError("Got impossible expression context: %r" % (node.ctx,))
 
     def CONTINUE(self, node):
@@ -1225,15 +1258,8 @@ class Checker(object):
         def runFunction():
 
             self.pushScope()
-            for name in args:
-                self.addBinding(node, Argument(name, node))
-            if isinstance(node.body, list):
-                # case for FunctionDefs
-                for stmt in node.body:
-                    self.handleNode(stmt, node)
-            else:
-                # case for Lambdas
-                self.handleNode(node.body, node)
+
+            self.handleChildren(node, omit='decorator_list')
 
             def checkUnusedAssignments():
                 """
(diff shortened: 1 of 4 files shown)


MacBox7:
This fixes the issue with minimum changes. The problem i.e being fixed here is in regard to the node to which the `Argument` node is being associated.

Previously the `source` attribute of `Argument` node i.e `node.source` pointed to `FunctionDef` node resulting in incorrect `lineno` and `column_offset` of the argument. But now the source attribute points to the `arg` node thus associating arguments with correct lineno and col_offset. This fix is only for PY3 as `arg` node was introduced in here. Patch has no changes for PY2.

MacBox7:
Just added some tests to make sure `ARG` and `ARGUMENT` handler are handled. With all possible permutations of the valid ways by which arguments can be supplied to the function.

pyflakes/test/test_other.py
```diff
@@ -1159,6 +1159,36 @@ def f(): global foo
         def g(): foo = 'anything'; foo.is_used()
         ''')
 
+    @skipIf(version_info < (3, 3), "Python >= 3.3 only")
```
jayvdb: Only the last one isnt legal syntax on py27

`def foo(a, b, c=0, *args, d=0, **kwargs)`

All the others are fine on versions before 3.3, and should not be skipped.

However, the bigger problem is that these tests do not verify the changes to pyflakes.

These tests can run on master, and they all pass, as they are in fact very similar to other functions we have elsewhere in the test suite.

To test your changes, tests need to

1. invoke `Checker` with a `scope` set to a value other than `ModuleScope` .  This can be done with `self.flakes( ... , scope=FunctionScope)`

2. *and* need to pass `Checker` the equivalent of `body[0]` in the issue https://github.com/PyCQA/pyflakes/issues/338 , but that cant be done with `self.flakes(..)` currently.  You might change the harness so that `input` can be either an `ast.Node`, or a string.

3. Inspect the deadScopes , similar to `test_scope_class`, to ensure it has the expected structure , on any version of Python, and any differences between Python versions are recorded into the test suite so they do not flippantly change in the future.

MacBox7:
New changes done to do PR:
1. Initially changes were only done for python3 arguments. Now more changes have been done to fix the arguments of python2 as well. The location information(line, col, node) in python2 now corresponds to ast.Name object for each respective argument. Initially, they were all pointing to `ast.FunctionDef` which was wrong and led to incorrect positional information. However `kwarg` and `vararg` still point to `ast.Function` as they are represented as plain string in python2.

2. Added a test harness that can be used to test code snippets rather than complete code.

MacBox7:
Need opinion on whether representing `kwarg` and `vararg` with `ast.argument` is a good choice.

pyflakes/test/test_code_segment.py
```diff
@@ -0,0 +1,139 @@
...
+        self.assertEqual(len(module_scopes), 0)
+        self.assertEqual(len(function_scopes), 2)
+
+        function_scope_foo = function_scopes[1]
+        function_scope_bar = function_scopes[0]
+
+        self.assertIsInstance(function_scope_foo, FunctionScope)
+        self.assertIsInstance(function_scope_bar, FunctionScope)
+
+        self.assertIn('a', function_scope_foo)
```
jayvdb: Need to assert which binding classes each of these things was.
(possibly assert all of them are the same, at least for `function_scope_bar`'s params

pyflakes/test/test_code_segment.py
```diff
@@ -0,0 +1,139 @@
...
+            def bar():
+                x = 0
+        '''
+        self.flakes(TestCodeSegments.parse_ast(code).body[0],
+                    m.UnusedVariable,
+                    scope=ClassScope)
+
+    def test_scope_class(self):
+        code = '''
+        class Foo:
```
jayvdb: add class variables

pyflakes/checker.py
```diff
@@ -1220,8 +1222,8 @@ def addArgs(arglist):
         def runFunction():
 
             self.pushScope()
-            for name in args:
-                self.addBinding(node, Argument(name, node))
+
+            self.handleNode(node.args, node)
             if isinstance(node.body, list):
```
jayvdb: Now you should also be able to replace this (and the above handleNode) with `self.handleChildren(node, omit='decorator_list')` .

the decorators are already processed at the beginning of the `FunctionDef` handler, as they are in the outer scope.

MacBox7: It is a single node. Why to use `handleChildren`?

pyflakes/test/harness.py
```diff
@@ -16,8 +16,13 @@ class TestCase(unittest.TestCase):
     withDoctest = False
 
     def flakes(self, input, *expectedOutputs, **kw):
-        tree = compile(textwrap.dedent(input), "<test>", "exec", PyCF_ONLY_AST)
-        w = checker.Checker(tree, withDoctest=self.withDoctest, **kw)
+        if kw.get('scope', None):
+            w = checker.Checker(input.body[0],
```
jayvdb: `tree = input.body[0]` then you do not need to modify the `w = checker.Checker(..)` line

pyflakes/checker.py
jayvdb: No, use `self.handleChildren(node, omit='decorator_list')` and replace the `if` block on *this* line.  The body of the node can now be routed through `handleChildren` because of the other improvements you have made.

pyflakes/checker.py
```diff
@@ -1252,6 +1258,19 @@ def checkReturnWithArgumentInsideGenerator():
 
         self.deferFunction(runFunction)
 
+    def ARGUMENTS(self, node):
+        parent_function = self.getAncestor(node, (ast.FunctionDef, ast.Lambda))
```
jayvdb: if you create a `getScopeNode` function, which gets the innermost functiondef or lambda or classdef or module , it should be the same result as what you want.

Your new `getScopeNode` can still be implemented as `self.getAncestor(node, (ast.FunctionDef, ast.Lambda, ast.ClassDef, ast.Module))` etc.  The wrapper `getScopeNode` just reduces the repetition, and supports any new scope which has args.

And this line should only occur in `if PY2:` as it is unneeded for Python 3

pyflakes/checker.py
```diff
@@ -481,7 +481,8 @@ class Checker(object):
     del _customBuiltIns
 
     def __init__(self, tree, filename='(none)', builtins=None,
-                 withDoctest='PYFLAKES_DOCTEST' in os.environ):
+                 withDoctest='PYFLAKES_DOCTEST' in os.environ,
+                 scope=ModuleScope):
```
jayvdb: do we even need this?

We could derive this from the `tree` 's `ast.Node` type.

Is there any scenario when the scope type shouldnt be derived from the `tree` type?
Less options is less to support.

pyflakes/checker.py
```diff
@@ -641,6 +642,15 @@ def descendantOf(self, node, ancestors, stop):
                 return True
         return False
 
+    def getAncestor(self, node, ancestor_type):
+        parent = node
+        while True:
+            if parent is self.root:
```
jayvdb: do you really need this?  Please check.

And due to subtree support, I think it is best to prefix this function name with `_` as the subtree support may need to be slightly adjusted as we learn more.

pyflakes/checker.py
```diff
@@ -97,7 +97,7 @@ def iter_child_nodes(node, omit=None, _fields_order=_FieldsOrder()):
     are nodes and all items of fields that are lists of nodes.
     """
     for name in _fields_order[node.__class__]:
-        if name == omit:
+        if omit and name in omit:
```
jayvdb: You need to support the old calling signature as well as the new one.

pyflakes/checker.py
MacBox7: Many more scopes could be added to pyflakes. For eg: a `TryScope`. Thus deriving from tree type might hard code this.

pyflakes/checker.py
```diff
@@ -1252,6 +1262,17 @@ def checkReturnWithArgumentInsideGenerator():
 
         self.deferFunction(runFunction)
 
+    def ARGUMENTS(self, node):
+        self.handleChildren(node, omit=('defaults', 'kw_defaults'))
+        if PY2:
+            if node.vararg:
+                self.addBinding(node, Argument(node.vararg, self.getScopeNode(node)))
```
jayvdb: please dont call `self.getScopeNode(node)` twice

pyflakes/checker.py
jayvdb: We worry about that then.

pyflakes/checker.py
jayvdb: hmm, it is, but only loosely because `'body' in 'body'` is True.  Anyway.. it works.
Add a note to the docstring that `omit` can be a single string or list of strings.

pyflakes/test/test_code_segment.py
MacBox7: Resolved

pyflakes/checker.py
```diff
@@ -507,6 +513,20 @@ def __init__(self, tree, filename='(none)', builtins=None,
         self.popScope()
         self.checkDeadScopes()
 
+    @staticmethod
+    def getScope(tree):
+        if isinstance(tree, ast.Module):
```
jayvdb: define a _private class mapping
```py
_ast_node_scope = {
    ast.Module: ModuleScope,
    ast.ClassDef: ClassScope,
    ast.FunctionDef: FunctionScope,
    ast.Lambda: FunctionScope,
    ...
}
```

Then this lookup is easy to do (and doesnt require new function), and `_ast_node_scope.keys()` can be used in `getScopeNode`.

PR #46, opened 2015-11-16
Report usage of star imports

jayvdb:

PR #343, opened 2018-07-16
Track locals()

jayvdb:


jayvdb:
Note this also corrects handling of `locals()` which only sees variables preceeding it, while the current approach marks all variables as used.  This is new test case `test_unusedVariableAfterLocals` in the patch.

pyflakes/checker.py
```diff
@@ -708,6 +706,15 @@ def handleNodeLoad(self, node):
...
         importStarred = None
 
+        if node.id == 'locals' and isinstance(node.parent, ast.Call):
+            # we are doing locals() call, which marks names currently
+            # in scope as used.
+            scope = self.scope
+            if isinstance(scope, GeneratorScope):
+                scope = self.scopeStack[-2]
+            for binding in scope.values():
+                binding.used = (self.scope, node)
```
jayvdb: this could be changed to

```
if not binding.used:
    binding.used = (self.scope, node)
```

Then at the end of the scope, any bindings used *only* by `locals()` could be a different type of message.

See https://github.com/PyCQA/pyflakes/pull/333#issuecomment-405147364

asmeurer:
IMHO `locals` is enough of a meta-programming technique that pyflakes should warn about it (same as `import *`).  Otherwise `locals()` breaks all unused variable checks that would occur before it is called. 

Having it warn but as a distinct error message that could be silenced by flake8 or whatever seems like the best solution.

asmeurer:
What exactly is the import * change you are referring to? Did that behavior change recently?

jayvdb:
No, I am referring to 1.1.0 change, https://github.com/PyCQA/pyflakes/pull/46/commits/0532189b34086740d7fe553bd87c0a6cf682a93b

PR #344, opened 2018-07-16
Simplify Python 3 except scope

jayvdb:
The existing variable with the same name given to the exception can only be in the same scope.

### DIFF
--- a/pyflakes/checker.py
+++ b/pyflakes/checker.py
@@ -1461,15 +1461,9 @@ class Checker(object):
         # to more accurately determine if the name is used in the except:
         # block.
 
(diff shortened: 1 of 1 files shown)


jayvdb:
ping @jdufresne

jdufresne:
Hi @jayvdb, I am not a maintainer of this project. I've only contributed small patches. Did you have a question directly for me?

jayvdb:
@jdufresne I believe you were involved in this chunk of code, and hope you may check this revision is correct.

jdufresne:
Looks fine to me.

PR #381, opened 2018-10-30
Don't leave behind temporary files when running the testsuite

asottile:
I noticed the testsuite left behind a bunch of temporary files on invocation, this cleans them up

### DIFF
--- a/pyflakes/test/test_api.py
+++ b/pyflakes/test/test_api.py
@@ -2,6 +2,7 @@
 Tests for L{pyflakes.scripts.pyflakes}.
 """
 
+import contextlib
 import os
 import sys
 import shutil
@@ -148,8 +149,7 @@ class TestIterSourceCode(TestCase):
     def makeEmptyFile(self, *parts):
         assert parts
         fpath = os.path.join(self.tempdir, *parts)
-        fd = open(fpath, 'a')
-        fd.close()
+        open(fpath, 'a').close()
         return fpath
 
     def test_emptyDirectory(self):
@@ -324,17 +324,20 @@ class CheckTests(TestCase):
     Tests for L{check} and L{checkPath} which check a file for flakes.
     """
 
+    @contextlib.contextmanager
     def makeTempFile(self, content):
         """
         Make a temporary file containing C{content} and return a path to it.
         """
-        _, fpath = tempfile.mkstemp()
-        if not hasattr(content, 'decode'):
-            content = content.encode('ascii')
-        fd = open(fpath, 'wb')
-        fd.write(content)
-        fd.close()
-        return fpath
+        fd, name = tempfile.mkstemp()
+        try:
+            with os.fdopen(fd, 'wb') as f:
+                if not hasattr(content, 'decode'):
+                    content = content.encode('ascii')
+                f.write(content)
+            yield name
+        finally:
+            os.remove(name)
 
     def assertHasErrors(self, path, errorList):
         """
@@ -372,8 +375,8 @@ class CheckTests(TestCase):
         exception to be raised nor an error indicator to be returned by
         L{check}.
         """
-        fName = self.makeTempFile("def foo():\n\tpass\n\t")
-        self.assertHasErrors(fName, [])
+        with self.makeTempFile("def foo():\n\tpass\n\t") as fName:
+            self.assertHasErrors(fName, [])
 
     def test_checkPathNonExisting(self):
         """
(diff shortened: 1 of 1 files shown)


asottile:
oops, windows, right...

asottile:
This is a little bit easier to review in no-whitespace mode:

https://github.com/PyCQA/pyflakes/pull/381/files?w=1

pyflakes/test/test_api.py
```diff
@@ -372,8 +375,8 @@ def test_missingTrailingNewline(self):
         exception to be raised nor an error indicator to be returned by
         L{check}.
         """
-        fName = self.makeTempFile("def foo():\n\tpass\n\t")
-        self.assertHasErrors(fName, [])
+        with self.makeTempFile("def foo():\n\tpass\n\t") as fName:
```
jayvdb: if you dont mind, when rebasing, could you also change the double quotes to single quotes on lines that you modify, where that is sensible.

It isnt a firm style policy of this project, but seems to be were the codebase is going.

or if you prefer, I can do the rebasing for you.

pyflakes/test/test_api.py
```diff
@@ -148,8 +149,7 @@ def tearDown(self):
     def makeEmptyFile(self, *parts):
         assert parts
         fpath = os.path.join(self.tempdir, *parts)
-        fd = open(fpath, 'a')
-        fd.close()
+        open(fpath, 'a').close()
```
jayvdb: this is new to me - why is this helpful?

pyflakes/test/test_api.py
sigmavirus24: If we just switch to `black` then we can stop nit-picking this kind of nonsense.

pyflakes/test/test_api.py
asottile: it's no different than the left hand side, but reduces the scope in which an unclosed file descriptor could leak (notice that I just removed the temporary variable)

pyflakes/test/test_api.py
asottile: or if you really want single quotes, you could use [`double-quote-string-fixer`](https://github.com/pre-commit/pre-commit-hooks)

pyflakes/test/test_api.py
jayvdb: im not sure that it changes possibility of leaks, but happy to merge anyway - its nice to read it this way.  I was just hoping to understand it better

asottile: it's more to prevent a future modification from inserting code between the two statements (an improvement of maintenance, not of functionality)

pyflakes/test/test_api.py
asottile: imo, unless you have an automated tool enforcing it (such as [pre-commit](https://pre-commit.com)) it's not worth bringing up or rebasing commits -- just me though :man_shrugging:

jayvdb: Firstly, please note rebasing is about a linear git history.  That was instituted for this project long before I became a regular contributor and maintainer, but I do appreciate the value of it vs the GitHub flow approaches.

Other than that, I agee. I am also of the opinion that automated tools enforcing rules is helpful, but not at the expense of massive rewrites for the sake of it.  I also run @coala , and that framework with the https://gitmate.io integration has the ability to lint only the modified lines according to repository rules, and report back on PRs to helpfully guide newcomers, including advising that a rebase is required, and providing a GitHub based command to do the rebase automatically, so then at least it is clear that the committer was an automation and didnt perform any additional changes to the author's commit.

It would be great to have clearer contributing guidelines and a properly enforced code of conduct, to make this project a bit more enjoyable to contribute in, and ideally add advanced tools to help, but by consensus of the existing contributors.  We have been a bit lax in that department.

Getting back to the rebasing thing, are you ok with having someone else as the committer?
It can be a bit offputting on GitHub as it show multiple avatars on the commit.  I like to be sure before I do that to someone elses PR.

asottile: I assume you'd just click `rebase and squash` when merging if you're trying to enforce a linear history -- this already causes "Authored by ... Committed by ..." so....

jayvdb: fwiw, I would rebase, and repush, to ensure the changes pass on top of the current master.  Sometimes it is easy to know that this is unnecessary, but it is better to do it systematically.  That was what the maintainers were before I was a committer, and I have continued to do.

And it is what many other orgs insist on, especially any on GitLab which makes that workflow seemless.

There are many other benefits of it, which are not so noticable inside GitHub.  But for example, the Squash and Merge creates a new commit, even for commits which are already on top of master.

Anyways, seems you are ok with it, so I'll proceed.

asottile: indeed, go for it!  personally I don't like rebase-squash because I can never know if my actual patch was merged or not but if that's how the project runs do your worst :)

asottile: also in this case you didn't even need to rebase my commit, since it gets squashed at merge

jayvdb: Ya, that is the downside, which is why @coala requires that the contributor does it themselves, or they invoke the bot to do it, but this project hasnt modernised.  But with more regular contributors, maybe we can change that. ;-)

asottile: alternatively you can just re-trigger the `/pr` CI jobs -- travis / appveyor will use the latest "merge" and then you don't need everyone to rebase every time you merge a PR

Issue #276, opened 2017-06-02
Incorrect F401/F822 results when `__all__` is defined from concatenated lists


edmorley:
Using Python 2.7.13 & pyflakes 1.5.0, and this testcase (extracted from [this file from isort](https://github.com/timothycrosley/isort/blob/0aca83884667f1b20605e3bbef48bacf4f8b77fa/isort/pie_slice.py)):
```python
import sys

PY2 = sys.version_info[0] == 2
PY3 = sys.version_info[0] == 3

common_exports = ['PY2', 'PY3', 'this_does_not_exist']

if PY2:
    from itertools import ifilter as filter
    __all__ = common_exports + ['filter', 'this_does_not_exist_either']
else:
    __all__ = common_exports
```

**Expected:**
* `F822 undefined name 'this_does_not_exist' in __all__`
* `F822 undefined name 'this_does_not_exist_either' in __all__`

**Actual:**
* No F822 errors displayed (when there should be two).
* A false positive `F401 'itertools.ifilter as filter' imported but unused`.

It appears that flake8 ignores the `__all__` export entirely if it is defined from concatenated lists. ie:
```python 
__all__ = [] + ['this_will_be_ignored']

__all__ = ['so_will_this'] + []
```

bitglue:
How could pyflakes correctly handle this case without being a Python interpreter?

edmorley:
> How could pyflakes correctly handle this case without being a Python interpreter?

For the specific testcase above (which I should have reduced a bit further) I agree it's a bit dubious.

However this reduced testcase demonstrates the issue too, and doesn't require pyflakes to be a Python interpreter:

pyflakes correctly reports the F822 in this:
```python
FOO = 'bar'

__all__ = ['FOO']
__all__ += ['this_does_not_exist']
```
(thanks to https://github.com/PyCQA/pyflakes/commit/f4c674a3a6e80c902147b5bb61ceb176dde0fd4a)

...however incorrectly doesn't find any errors in:
```python
FOO = 'bar'

__all__ = ['FOO'] + ['this_does_not_exist']
```

bitglue:
Does anyone ever actually do `__all__ = ['FOO'] + ['this_does_not_exist']`?

edmorley:
Perhaps not.

[...]

PR #382, opened 2018-10-31
checker.py: Add support for list concatenation

CLiu13:
This adds support for when `'__all__'`
is defined from concatenated lists.
However, because pyflakes is not an
interpreter, this only applies to
certain reduced test cases.

Fixes https://github.com/PyCQA/pyflakes/issues/276

### DIFF
--- a/pyflakes/checker.py
+++ b/pyflakes/checker.py
@@ -390,10 +390,10 @@ class ExportBinding(Binding):
     can be determined statically, they will be treated as names for export and
     additional checking applied to them.
 
-    The only C{__all__} assignment that can be recognized is one which takes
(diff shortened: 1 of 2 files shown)


pyflakes/checker.py
```diff
@@ -408,6 +408,28 @@ def __init__(self, name, source, scope):
             for node in source.value.elts:
                 if isinstance(node, ast.Str):
                     self.names.append(node.s)
+        # If concatenating lists
+        elif isinstance(source.value, ast.BinOp):
```
jayvdb: I think you can use `ast.literal_eval` for this

Issue #280, opened 2017-06-19
Unreported undefined name in __all__


adhikasp:
Pyflakes was unable to detect undefined name in `__all__` when there exist undefined name outside of it. This make me need to use pyflakes 2 times to really detect all the errors. 

For example, given this code...

```python
from math import * # contain sin, cos, and csc
__all__ = ['sin', 'cos']
csc(1)
```

Pyflakes will report 

```
test.py:1: 'from math import *' used; unable to detect undefined names
test.py:5: 'csc' may be undefined, or defined from star imports: math
```

According to that, the code just have 2 "error", which is usage of `*` import and `csc` is undefined.
If I then fix the code 

```python
from math import csc
__all__ = ['sin', 'cos']
csc(1)
```

Then additional error shown up

```
test.py:3: undefined name 'cos' in __all__
test.py:3: undefined name 'sin' in __all__
```

Which then could finally fixed by

```python
from math import csc, cos, sin
__all__ = ['sin', 'cos']
csc(1)
```

bitglue:
What behavior do you expect? `import *` could import anything, so what could pyflakes do better than what it does now?

asmeurer:
He is saying there should be similar messages for sin and cos as the csc message.

adhikasp:
@bitglue I was expecting that if pyflakes do want to report about that 2 kind of error, do it at the same time, not when the other issue was already present (issue hidden at first sight)

bitglue:
It's "hidden" because pyflakes can't know what `*` imports. It's not until you explicitly declare what's in the namespace that pyflakes can tell you what's not in it.

PR #385, opened 2018-11-03
checker.py: Report undefined names in __all__

CLiu13:
This allows pyflakes to report undefined names
in `__all__` when `import *` is used, without having
to use pyflakes 2 times to detect the errors.

Fixes https://github.com/PyCQA/pyflakes/issues/280

### DIFF
--- a/pyflakes/checker.py
+++ b/pyflakes/checker.py
@@ -665,9 +665,16 @@ class Checker(object):
 
                 # mark all import '*' as used by the undefined in __all__
                 if scope.importStarred:
+                    from_list = []
                     for binding in scope.values():
                         if isinstance(binding, StarImportation):
                             binding.used = all_binding
+                            from_list.append(binding.fullName)
+                    # report * usage, with a list of possible sources
+                    from_list = ', '.join(sorted(from_list))
+                    for name in undefined:
+                        self.report(messages.ImportStarUsage,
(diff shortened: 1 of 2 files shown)


asmeurer:
This change seems odd. The names sin and cos *are* defined by the `import *`, so the error message is effectively wrong.

bitglue:
Won't this be a problem for people who want to use `import *` and `__all__`? Are we just saying, "if you use pyflakes, don't use `import *` ever?"

asmeurer:
Isn't that already the case? Pyflakes has always unconditionally warned about import *, even in `__init__.py`.

asmeurer:
IMHO, if you are using `__all__`, you should not be using `import *`, because you won't be warned about unused imports either from forgetting to put something in `__all__` or mispelling something. Pyflakes forces you to write things twice, once in the import and once in `__all__`, but I actually find this "double entry bookkeeping" useful. 

The thing that's personally annoying to me is that you can't use `import *` at all, even in cases where it can be legitimate. For instance, in an `__init__.py` file you could have

```py
from .submod import *
```

where `submod` uses `__all__`. Explicitly redefining `__all__` in the top-level `__init__.py` is useless and actually doing so would probably lead to bugs rather than prevent them, due to the ease of forgetting to add new names to both places.

Issue #366, opened 2018-09-25
Exclude "import ... as _" from unused-imports


anntzer:
Similarly to #202/#335 (excluding `_` from unused-variable checks), it would be appreciated if `import ... as _` was excluded from unused-import checks.

An example use case is when the module is imported solely for side-effects, see e.g. https://github.com/matplotlib/matplotlib/pull/12249.

Example repro (pyflakes 2.0 and master):
```
$ python -mpyflakes <(printf 'from foo import bar as _\n')
/proc/self/fd/11:1: 'foo.bar as _' imported but unused
```

graingert:
I tend to use `importlib.import_module("path.to.side.effect.module")` when requiring side effects

anntzer:
IMHO that just seems like "writing less natural code just to make the linter happy"?

sigmavirus24:
Is it really? In my opinion, it's a clearer signal of "We're importing this for some other reason." There's actually a convention of importing something from an internationalization (i18n) library as `_` in many large projects and then calling `_("test to be translated")`.

graingert:
importlib.import_module also has the benefit of not being moved my isort

anntzer:
Fair point re: collision with gettext (well, I haven't ever used it, but I can guess the idea).
OTOH if someone just writes `importlib.import_module("foo")` at the toplevel I don't think it would be particularly clear that they're intentionally doing an import for side-effects -- at least to me `import foo as _` is clearer given that binding to `_` is a standard convention for "do something and ignore its results".
In any case, I think that even if there are alternatives (i.e. importlib.import_module) available, pyflakes *could* still support `import ... as _`?

sigmavirus24:
If we "support" that, then people importing any i18n library and aliasing it as `_` will not be warned when they're no longer using it in that module which seems bad

anntzer:
But gettext even supports `gettext.install` (https://docs.python.org/3/library/gettext.html#gettext.install) which patches *builtins* with `builtins._` and there's no way for pyflakes to detect whether `_` is used at all.
Or you could also special-case `from gettext import _` and handle that import as a "normal" one.

mgedmin:
In many projects I've imported gettext-equivalent translation functions as `_`.  I appreciate flake8 telling me when those are no longer used and can be removed.

To me the canonical spelling of "this import used for side effects only" is
```python
import something  # noqa: does thins thing as a side effect that we need done here
```

graingert:
another option:

```python
[...]

PR #391, opened 2018-11-22
Mark Importation with underscore alias as used

seeeturtle:
If the importation has underscore alias like:
`from fu import bar as _`
it should be ignored from checking.

Closes https://github.com/PyCQA/pyflakes/issues/366

### DIFF
--- a/pyflakes/checker.py
+++ b/pyflakes/checker.py
@@ -785,7 +785,7 @@ class Checker(object):
 
         if value.name in self.scope:
             # then assume the rebound name is used as a global or within a loop
-            value.used = self.scope[value.name].used
+            value.used = value.used or self.scope[value.name].used
 
         self.scope[value.name] = value
 
@@ -1429,6 +1429,10 @@ class Checker(object):
             else:
                 name = alias.asname or alias.name
                 importation = Importation(name, node, alias.name)
+
+            if alias.asname == '_':
+                importation.used = True
+
             self.addBinding(node, importation)
 
     def IMPORTFROM(self, node):
@@ -1461,6 +1465,10 @@ class Checker(object):
             else:
                 importation = ImportationFrom(name, node,
                                               module, alias.name)
+
+            if alias.asname == '_':
+                importation.used = True
+
             self.addBinding(node, importation)
 
     def TRY(self, node):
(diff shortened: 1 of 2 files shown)


asmeurer:
I thought it was already conceded that `_` should not be marked as unused in other contexts.

bitglue:
Hmm, I guess it was. https://github.com/PyCQA/pyflakes/issues/202#issuecomment-386648451 seemed one of the better reasoned arguments, though now the [google style guide discourages using `_` for unused variables](https://google.github.io/styleguide/pyguide.html#214-decision).

> Unused argument warnings can be suppressed by deleting the variables at the beginning of the function. Always include a comment explaining why you are deleting it. “Unused.” is sufficient. For example:

> ```
> def viking_cafe_order(spam, beans, eggs=None):
>     del beans, eggs  # Unused by vikings.
>     return spam + spam + spam
> ```

> Other common forms of suppressing this warning include using `_` as the identifier for the unused argument, prefixing the argument name with `unused_`, or assigning them to `_`. These forms are allowed but no longer encouraged. The first two break callers that pass arguments by name, while the last does not enforce that the arguments are actually unused.

Ironic, but I guess it's not all that important of a point.

asmeurer:
Well unlike the `_ = ` case, which I still think is stupid since you can always just remove the assignment,  imports always require assigning a variable (similar to loop variables). And there are legitimate use cases where you want to import something without "using" it. Using `import as _` seems like a clean way to silence errors (I personally use something like `import module; module`, but maybe `import module as _` is cleaner).

bitglue:
`import module; del module` works to silence the error too, and ensures the module really is unused.

bitglue:
Thanks for the patch, but I'm rejecting it.

Largely this is because I saw https://github.com/PyCQA/pyflakes/issues/615, and then I reviewed the old issues about `_`, and came to believe any kind of "I don't mean to use this" naming convention in Pyflakes is probably a mistake.

It's a mistake because:

1. Everyone will want a different convention. Some people want `_`. Some want `__`. Some want anything starting with `_`. But some explicitly _don't_ want anything starting with `_`, because they use that prefix to mean something else. Some don't want `_` because they use that for gettext. There is no answer that works for everyone, and all possible answers harm someone.
2. The Google style guide, which [won the argument about `_` previously](https://github.com/PyCQA/pyflakes/issues/202#issuecomment-386799649), [no longer advocates for this usage](https://github.com/PyCQA/pyflakes/pull/391#issuecomment-469430086). So I don't even think we can say there's a prevailing convention.
3. `del` offers a better solution. It's explicit. It already works. It communicates the intent to the linter _and_ the runtime, and ensures the variable is actually unused, as intended.

PR #396, opened 2018-12-31
support module scope __annotations__

### LINKED ISSUE #395: undefined name '__annotations__' when use __annotations__ in module level
Environment:
- Python==3.7.1
- pyflakes==2.0.0

When using pyflake to check following code:

```Python3
# t.py
a: int = 2
print(__annotations__)
```

It will reported that

```
t.py:3: undefined name '__annotations__'
```
About `__annotations__`, please see [PEP 526 -- Syntax for Variable Annotations](https://www.python.org/dev/peps/pep-0526/)
> In addition, at the module or class level, if the item being annotated is a simple name, then it and the annotation will be stored in the `__annotations__` attribute of that module or class (mangled if private) as an ordered mapping from names to evaluated annotations. 

And `__annotations__` is not a [`builtin_vars`](https://github.com/PyCQA/pyflakes/blob/master/pyflakes/checker.py#L21), so it will not found.

I think this attribute should be added in [`_MAGIC_GLOBALS = ['__file__', '__builtins__', 'WindowsError']`](https://github.com/PyCQA/pyflakes/blob/master/pyflakes/checker.py#L488)

If you accept this solution, I will pull request to fix it :)

Hanaasagi:
Fixes #395

### DIFF
--- a/pyflakes/checker.py
+++ b/pyflakes/checker.py
@@ -12,6 +12,7 @@ import sys
 
 PY2 = sys.version_info < (3, 0)
 PY34 = sys.version_info < (3, 5)    # Python 2.7 to 3.4
+PY35 = sys.version_info < (3, 6)    # Python 2.7 to 3.5
 try:
     sys.pypy_version_info
     PYPY = True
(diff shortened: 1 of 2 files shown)


sigmavirus24:
@asottile that was my thought on this. It's not @Hanaasagi 's problem but those variables are pretty terrible

PR #397, opened 2018-12-31
Add python and platform details to --version

### LINKED ISSUE #283: `pyflakes --version` should indicate the version of the interpreter as well
At the moment if someone provides

```
$ pyflakes --version
1.5.0
```

As output, there's no easy way to know what version of Python they've installed PyFlakes onto. Let's also include something like `flake8 --version`

```
3.2.1 (pyflakes: 1.3.0, flake8-docstrings: 1.0.2, pydocstyle: 1.1.1, pycodestyle: 2.2.0, mccabe: 0.5.3) CPython 2.7.13 on Linux
```

We don't need to necessarily worry about plugins but having the interpreter, its version, and the OS would be useful debugging information.

addisonElliott:
Includes Python version & OS details to `--version` output. 

Useful since pyflakes output changes depending on what version of Python the package is ran for.

Format follows `flake8` commit to include additional information here: [https://gitlab.com/pycqa/flake8/commit/c9fb680dffa37517bbda76cc2b572d6f192508bd](https://gitlab.com/pycqa/flake8/commit/c9fb680dffa37517bbda76cc2b572d6f192508bd)

Fixes #283

### DIFF
--- a/pyflakes/api.py
+++ b/pyflakes/api.py
@@ -5,6 +5,7 @@ from __future__ import with_statement
 
 import sys
 import os
+import platform
 import re
 import _ast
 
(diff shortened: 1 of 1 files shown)


sigmavirus24:
This looks good to me. Then again I authored https://gitlab.com/pycqa/flake8/commit/c9fb680dffa37517bbda76cc2b572d6f192508bd so I probably shouldn't review this too =P

PR #398, opened 2018-12-31
Update TravisCI to use flake8 3.6.0

addisonElliott:
Update TravisCI & `tox.ini` to use v3.6.0 in flake8 rather than v2.1.0 which is 4 years old.

Update `tox.ini` to ignore W504. This is ignored by default in flake8 but it is enabled by the `select`. *W503* and *W504* are opposites of one another and determines when to break with a statement and an operator.  PEP8 recommends breaking **after** binary operator (they originally recommended breaking before but changed).

Fix some expressions that break before binary operators.

Move import statement to top to fix another warning.

Change pep8 to pycodestyle and bump version

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,4 +18,4 @@ install:
   - pip list
 script:
   - python setup.py test -q
-  - if [ "$TRAVIS_PYTHON_VERSION" != "nightly" ]; then pip install flake8==2.1.0 pep8==1.5.6 && flake8 --version && flake8 pyflakes setup.py; fi
+  - if [ "$TRAVIS_PYTHON_VERSION" != "nightly" ]; then pip install flake8==3.6.0 pep8==1.5.6 && flake8 --version && flake8 pyflakes setup.py; fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,4 +18,4 @@ install:
   - pip list
 script:
   - python setup.py test -q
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -18,4 +18,4 @@ install:
   - pip list
 script:
   - python setup.py test -q
-  - if [ "$TRAVIS_PYTHON_VERSION" != "nightly" ]; then pip install flake8==2.1.0 pep8==1.5.6 && flake8 --version && flake8 pyflakes setup.py; fi
+  - if [ "$TRAVIS_PYTHON_VERSION" != "nightly" ]; then pip install flake8==3.6.0 pycodestyle==2.4.0 && flake8 --version && flake8 pyflakes setup.py; fi
```
sigmavirus24: Flake8 manages its own pycodestyle version, why have it here?

tox.ini
```diff
@@ -6,12 +6,13 @@ envlist =
 [testenv]
 deps =
```
sigmavirus24: flake8 should be listed here

tox.ini
```diff
@@ -6,12 +6,13 @@ envlist =
...
 commands =
-    pip install flake8==2.1.0 pep8==1.5.6
+    pip install flake8==3.6.0 pycodestyle==2.4.0
```
sigmavirus24: There's no need to pin pycodestyle here and the installation should happen in `deps = flake8`

addisonElliott:
@sigmavirus24 Agree with your suggestions & I just changed them. I had just bumped the versions without thinking much about whether they were necessary or not.

PR #400, opened 2019-01-01
Parse PEP 484 type comments

### LINKED ISSUE #247: Detect type hint comments: PEP484
*Original report by [w-thomas-q](https://launchpad.net/~w-thomas-q) on [Launchpad](https://bugs.launchpad.net/bugs/1587170):*

Hi,

thanks for this really great tool!
One issues I stumbled upon was that type hints seem to be ignored / not detected.
For example, we have an import at the top of a file:

    from typing import List

And then later in an `__init__` function something like this:

    self.scripts = []  # type: List[script.Script]

Now running pyflakes (through flake8) gives me this error:

    foo.py:4:1: F401 'List' imported but unused

PEP484 defines this kind of type comments.
Is there an option to enable type hint detection?
Or can we have that as feature request? I guess I'm fine with type comments only for the time being, but of course a fully PEP484 compliant solution would be greatly appreciated :)

Thanks!

asottile:
Resolves #247

### DIFF
--- a/pyflakes/checker.py
+++ b/pyflakes/checker.py
@@ -6,9 +6,14 @@ Also, it models the Bindings and Scopes.
 """
 import __future__
 import ast
+import bisect
+import collections
 import doctest
+import functools
 import os
+import re
 import sys
+import tokenize
 
 from pyflakes import messages
 
@@ -23,6 +28,10 @@ except AttributeError:
 
 builtin_vars = dir(__import__('__builtin__' if PY2 else 'builtins'))
 
+if PY2:
+    tokenize_tokenize = tokenize.generate_tokens
+else:
+    tokenize_tokenize = tokenize.tokenize
 
 if PY2:
     def getNodeType(node_class):
@@ -62,6 +71,13 @@ if PY35_PLUS:
 else:
     FOR_TYPES = (ast.For,)
     LOOP_TYPES = (ast.While, ast.For)
+
+# https://github.com/python/typed_ast/blob/55420396/ast27/Parser/tokenizer.c#L102-L104
+TYPE_COMMENT_RE = re.compile(r'^#\s*type:\s*')
+# https://github.com/python/typed_ast/blob/55420396/ast27/Parser/tokenizer.c#L1400
+TYPE_IGNORE_RE = re.compile(TYPE_COMMENT_RE.pattern + r'ignore\s*(#|$)')
+# https://github.com/python/typed_ast/blob/55420396/ast27/Grammar/Grammar#L147
+TYPE_FUNC_RE = re.compile(r'^(\(.*?\))\s*->\s*(.*)$')
 
 
 class _FieldsOrder(dict):
@@ -522,6 +538,63 @@ def is_typing_overload(value, scope):
     )
 
 
+def make_tokens(code):
+    # PY3: tokenize.tokenize requires readline of bytes
+    if not isinstance(code, bytes):
+        code = code.encode('UTF-8')
+    lines = iter(code.splitlines(True))
+    # next(lines, b'') is to prevent an error in pypy3
+    return tuple(tokenize_tokenize(lambda: next(lines, b'')))
+
+
+class _TypeableVisitor(ast.NodeVisitor):
+    """Collect the line number and nodes which are deemed typeable by
+    PEP 484
+
+    https://www.python.org/dev/peps/pep-0484/#type-comments
+    """
+    def __init__(self):
+        self.typeable_lines = []  # type: List[int]
+        self.typeable_nodes = {}  # type: Dict[int, ast.AST]
+
+    def _typeable(self, node):
+        # if there is more than one typeable thing on a line last one wins
+        self.typeable_lines.append(node.lineno)
+        self.typeable_nodes[node.lineno] = node
+
+        self.generic_visit(node)
+
+    visit_Assign = visit_For = visit_FunctionDef = visit_With = _typeable
+    visit_AsyncFor = visit_AsyncFunctionDef = visit_AsyncWith = _typeable
+
+
+def _collect_type_comments(tree, tokens):
+    visitor = _TypeableVisitor()
+    visitor.visit(tree)
+
+    type_comments = collections.defaultdict(list)
+    for tp, text, start, _, _ in tokens:
+        if (
+                tp != tokenize.COMMENT or  # skip non comments
+                not TYPE_COMMENT_RE.match(text) or  # skip non-type comments
+                TYPE_IGNORE_RE.match(text)  # skip ignores
+        ):
+            continue
+
+        # search for the typeable node at or before the line number of the
+        # type comment.
+        # if the bisection insertion point is before any nodes this is an
+        # invalid type comment which is ignored.
+        lineno, _ = start
+        idx = bisect.bisect_right(visitor.typeable_lines, lineno)
+        if idx == 0:
+            continue
+        node = visitor.typeable_nodes[visitor.typeable_lines[idx - 1]]
+        type_comments[node].append((start, text))
+
+    return type_comments
+
+
 class Checker(object):
     """
     I check the cleanliness and sanity of Python code.
@@ -556,8 +629,11 @@ class Checker(object):
         builtIns.update(_customBuiltIns.split(','))
     del _customBuiltIns
 
+    # TODO: tokens= is required to perform checks on type comments, eventually
+    #       make this a required positional argument.  For now it is defaulted
+    #       to `()` for api compatibility.
     def __init__(self, tree, filename='(none)', builtins=None,
-                 withDoctest='PYFLAKES_DOCTEST' in os.environ):
+                 withDoctest='PYFLAKES_DOCTEST' in os.environ, tokens=()):
         self._nodeHandlers = {}
         self._deferredFunctions = []
         self._deferredAssignments = []
@@ -573,6 +649,7 @@ class Checker(object):
             raise RuntimeError('No scope implemented for the node %r' % tree)
         self.exceptHandlers = [()]
         self.root = tree
+        self._type_comments = _collect_type_comments(tree, tokens)
         for builtin in self.builtIns:
             self.addBinding(None, Builtin(builtin))
         self.handleChildren(tree)
@@ -952,7 +1029,26 @@ class Checker(object):
             except KeyError:
                 self.report(messages.UndefinedName, node, name)
 
+    def _handle_type_comments(self, node):
+        for (lineno, col_offset), comment in self._type_comments.get(node, ()):
+            comment = comment.split(':', 1)[1].strip()
+            func_match = TYPE_FUNC_RE.match(comment)
+            if func_match:
+                parts = (func_match.group(1), func_match.group(2).strip())
+            else:
+                parts = (comment,)
+
+            for part in parts:
+                if PY2:
+                    part = part.replace('...', 'Ellipsis')
+                self.deferFunction(functools.partial(
+                    self.handleStringAnnotation,
+                    part, node, lineno, col_offset,
+                    messages.CommentAnnotationSyntaxError,
+                ))
+
     def handleChildren(self, tree, omit=None):
+        self._handle_type_comments(tree)
         for node in iter_child_nodes(tree, omit=omit):
             self.handleNode(node, tree)
 
@@ -1040,7 +1136,7 @@ class Checker(object):
         self.addBinding(None, Builtin('_'))
         for example in examples:
             try:
-                tree = compile(example.source, "<doctest>", "exec", ast.PyCF_ONLY_AST)
+                tree = ast.parse(example.source, "<doctest>")
             except SyntaxError:
                 e = sys.exc_info()[1]
                 if PYPY:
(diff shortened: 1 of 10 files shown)


asottile:
some of the patches here are a little unrelated, let me know and I can split them out into separate PRs to focus on the changes here:
- `setup.py test` -> `-m unittest discovery pyflakes`
    - I had some major problems getting flake8 to install with `pyflakes.egg-info` sitting around from the `setup.py test` invocations.  after resolving those, I _think_ this was unnecessary but imo an improvement
- `compile(4 arg)` -> `ast.parse(...)` -- completely unrelated, but simplified some things
- modern flake8 - I'm using a not-yet-released version which contains [this patch](https://gitlab.com/pycqa/flake8/commit/a1532441917ab9ea70838903095ffe1d802e541c)

pyflakes/checker.py
```diff
@@ -518,6 +527,60 @@ def is_typing_overload_decorator(node):
...
+
+
+class _TypeableVisitor(ast.NodeVisitor):
+    """Collect the line number and nodes which are deemed typeable by
+    PEP 484
+
+    https://www.python.org/dev/peps/pep-0484/#type-comments
+    """
+    def __init__(self):
+        self.typeable_lines = []  # type: int
```
sigmavirus24: Should tihs be `# type: List[int]`?

pyflakes/checker.py
```diff
@@ -518,6 +527,60 @@ def is_typing_overload_decorator(node):
...
+        self.typeable_nodes = {}  # type: Dict[int, ast.AST]
+
+    def _typeable(self, node):
+        # if there is more than one typeable thing on a line last one wins
+        self.typeable_lines.append(node.lineno)
```
sigmavirus24: Perhaps this is silly, but why are we doubly tracking the line number? Could `typeable_lines` simply be a property that does `list(self.typeable_nodes.values())`?

pyflakes/checker.py
```diff
@@ -518,6 +527,60 @@ def is_typing_overload_decorator(node):
...
+def _collect_type_comments(tree, tokens):
+    visitor = _TypeableVisitor()
+    visitor.visit(tree)
+
+    type_comments = collections.defaultdict(list)
+    for tp, text, start, _, _ in tokens:
+        if (
+                tp != tokenize.COMMENT or  # skip non comments
+                not text.startswith('# type:') or  # skip non-type comments
+                text.split()[:3] == ['#', 'type:', 'ignore']  # skip ignores
```
sigmavirus24: Perhaps another silly question, why are we doing `text.split()[:3]` to check this but not using split above to see if the comment starts with `# type:`?

My guess is that the split here is to accommodate multiple spaces so that all of the following work:

```
# type: ignore
#   type:       ignore
# type:    ignore
#      type: ignore
```

But then we're going to bail on those lines anyway because of `not text.startswith('# type:')` so I'd like to understand the decision here. Do we perhaps want to do:

```py
for tp, text, start, _, _ in tokens:
    text_pieces = text.split()
    if (
        tp != tokenize.COMMENT or
        not (text_pieces[:2] == ['#', 'type:'] and
             text_pieces[3] == 'ignore')
    ):
        comment
```

sigmavirus24: I don't know the intricacies of mypy but does it support `#type:ignore` as well? Do we need to handle that? Should this be a regular expression?

pyflakes/checker.py
```diff
@@ -948,7 +1013,25 @@ def on_conditional_branch():
             except KeyError:
                 self.report(messages.UndefinedName, node, name)
 
+    def _handle_type_comments(self, node):
+        for (lineno, col_offset), comment in self._type_comments.get(node, ()):
+            comment = comment[len('# type:'):].strip()
+            func_match = re.match(r'^(\(.*?\)) *-> *(.*)$', comment)
```
sigmavirus24: If I remember correctly, compiling the regular expression can be expensive. Should we do that once at the module level and re-use it over-and-over instead of compiling each time?

pyflakes/checker.py
asottile: yes, yes it should -- I should write less code while celebrating new years :laughing:

pyflakes/checker.py
asottile: I couldn't think of a way to make the lookup efficient without needing two state variables here.  having a property which sorts the keys would invoke a O(N) (`list(typeable_nodes)`) + O(N*log(N)) (`sorted(...)`) + O(log(N)) (`bisect.bisect_right(...)`) instead of just the `O(log(N))` lookup

pyflakes/checker.py
asottile: yeah looks like there's more to support here, I'll make a regex for this.  Even this is valid: `#type:()->None`

pyflakes/checker.py
asottile: you are correct, I thought I had hoisted these but looks like I forgot one!

pyflakes/checker.py
sigmavirus24: Makes sense. :+1:

pyflakes/checker.py
```diff
@@ -552,7 +622,8 @@ class Checker(object):
         builtIns.update(_customBuiltIns.split(','))
     del _customBuiltIns
 
-    def __init__(self, tree, filename='(none)', builtins=None,
+    # TODO: tokens=() is just to make flake8 happy
+    def __init__(self, tree, tokens=(), filename='(none)', builtins=None,
```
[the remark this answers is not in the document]
asottile: if I move `tokens=()` the api is consistent, though what I really would want is `tokens` to be a required argument.  that said, if it is left off pyflakes will still lint -- just it won't check type comments

pyflakes/test/test_checker.py
```diff
@@ -0,0 +1,182 @@
...
+        self.assertSetEqual(ret, set())
+
+    def test_type_ignore_comment_ignored(self):
+        """
+        Test that `# type: ignore` comments are not collected.
+        """
+        ret = self._collect('x = 1  # type: ignore')
+        self.assertSetEqual(ret, set())
+
+    def test_type_ignore_with_other_things_ignored(self):
```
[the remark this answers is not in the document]
asottile: will do, it's not exactly an interesting case from the implementation's standpoint though -- the only reason this is here was to regress-test the edge-case of `ignore` parsing -- I can add this nonetheless

pyflakes/test/test_other.py
```diff
@@ -2073,6 +2073,45 @@ class A:
...
+
+        x = 1  #type:F
+        """)
+
+    def test_typeCommentsInvalidDoesNotMarkAsUsed(self):
+        self.flakes("""
+        from mod import F
+
+        # type: F
+        """, m.UnusedImport)
```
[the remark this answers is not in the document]
asottile: yep this is intentional

pyflakes/checker.py
```diff
@@ -518,6 +534,60 @@ def is_typing_overload_decorator(node):
...
+    for tp, text, start, _, _ in tokens:
+        if (
+                tp != tokenize.COMMENT or  # skip non comments
+                not TYPE_COMMENT_RE.match(text) or  # skip non-type comments
+                TYPE_IGNORE_RE.match(text)  # skip ignores
+        ):
+            continue
+
+        lineno, _ = start
+        idx = bisect.bisect_right(visitor.typeable_lines, lineno)
```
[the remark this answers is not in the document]
asottile: my goal with "associate to previous" was to get the proper scope in which it has an effect such that pyflakes can check the right variables there.  you're right this example doesn't quite make sense and that `typed-ast` errors on this.  I couldn't quite make this happen with just cross referencing.

The actual reason it's "look backwards" is for functions which are annotated as follows:

```python
def f(
    a,  # type: int  # (annotation for a specific argument)
):
    # type: (...) -> None # (annotation for the function's return value)
    pass
```

In this case it's necessary to assign both the arguments and the type signature to the function itself, and make sure to evaluate those in the module scope instead of the function scope

pyflakes/test/test_checker.py
```diff
@@ -0,0 +1,182 @@
...
+
+    def test_last_node_wins(self):
+        """
+        Test that when two typeable nodes are present on a line, the last
+        typeable one wins.
+        """
+        ret = self._collect('def f(): x = 1  # type: int')
+        self.assertSetEqual(ret, {(ast.Assign, ('# type: int',))})
+
+    def test_function_def_assigned_comments(self):
```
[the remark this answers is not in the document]
asottile: It should work, not by accident (I can add a test for this)

`(str, List[int])` is a tuple expression, which itself is an expression statement

pyflakes/checker.py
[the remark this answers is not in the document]
asottile: this'll require two branches instead of 1 -- kind of a leaky abstraction.  I'll add a comment instead if it's not clear

pyflakes/test/test_other.py
```diff
@@ -2073,6 +2073,45 @@ class A:
         class B: pass
         ''', m.UndefinedName)
 
+    def test_typeCommentsMarkImportsAsUsed(self):
```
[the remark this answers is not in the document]
asottile: hah oh boy this class got abused for all sorts of things -- I just stuck this next to the other type annotation tests.

How about I split this up in a followup? there's a lot of diff churn already here and I think it would make sense for just a "mv" commit that's easy to review

pyflakes/checker.py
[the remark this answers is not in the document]
asottile: pyflakes has an `api.py` file which I suspect _should_ be the public api, however even flake8 reaches into pyflakes/checker.py :man_facepalming:

pyflakes/test/test_other.py
```diff
@@ -2073,6 +2073,45 @@ class A:
...
+        from mod import F
+
+        # type: F
+        """, m.UnusedImport)
+
+    def test_typeCommentsSyntaxError(self):
+        self.flakes("""
+        def f(x):  # type: (F[) -> None
+            pass
+        """, m.ForwardAnnotationSyntaxError)
```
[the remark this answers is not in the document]
asottile: urgh but then I need to thread that through -- I'll do this but it breaks the abstraction

pyflakes/checker.py
```diff
@@ -552,7 +622,8 @@ class Checker(object):
         builtIns.update(_customBuiltIns.split(','))
     del _customBuiltIns
 
-    def __init__(self, tree, filename='(none)', builtins=None,
+    # TODO: tokens=() is just to make flake8 happy
```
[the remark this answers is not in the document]
asottile: yeah this is api shape issue -- I'm a bit stuck here on the right thing to do. Here's what I want and the constraints:

- tree + tokens are required position arguments -- they are necessary to perform all of the checks as implemented in this PR
- flake8 <=3.6.0 calls `Checker` directly with positional arguments (in 3.7.0+ it will call it with tree positionally and `filename=` as a named argument)
- pyflakes uses flake8 to validate itself

I *think* the easiest way forward is to soft-break the api for those which are calling `Checker()` without named arguments and release a minor version with `tokens=()`.  flake8 pins a specific version of pyflakes and can therefore update in lockstep safely.  At a later date, pyfalkes makes `tokens` a required positional argument and releases a major release.

Alternatively, we could just hard-break today and I could change the tests to create separate flake8 environments which don't use the source tree's pyflakes to lint things -- though this is much messier than what I have right now.

I think for option 1, the branch as-is is fine, though I should update the comment

pyflakes/checker.py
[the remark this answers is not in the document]
asottile: going to resolve this discussion -- let's continue it in the one where I posted the big blob about flake8

pyflakes/test/test_other.py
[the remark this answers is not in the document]
asottile: Created followup issue here: https://github.com/PyCQA/pyflakes/issues/402

pyflakes/checker.py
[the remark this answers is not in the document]
asottile: I'm adding a test and comment comment demonstrating and explaining this limitation.

pyflakes/checker.py
```diff
@@ -518,6 +534,63 @@ def is_typing_overload_decorator(node):
...
+
+def _collect_type_comments(tree, tokens):
+    visitor = _TypeableVisitor()
+    visitor.visit(tree)
+
+    type_comments = collections.defaultdict(list)
+    for tp, text, start, _, _ in tokens:
+        if (
+                tp != tokenize.COMMENT or  # skip non comments
+                not TYPE_COMMENT_RE.match(text) or  # skip non-type comments
```
sigmavirus24: Do we want match or search? I wonder, can mypy/typed_ast handle something like:


```py
foo = 1   # noqa: E11  # type: ignore
```

asottile: typed-ast is picky and requires the type comment to come first:

(using [astpretty](https://github.com/asottile/astpretty))

```
$ cat t.py 
x = 1  # noqa  # type: int
$ astpretty --typed-27 t.py 
Module(
    body=[
        Assign(
            lineno=1,
            col_offset=0,
            targets=[Name(lineno=1, col_offset=0, id='x', ctx=Store())],
            value=Num(lineno=1, col_offset=4, n=1),
            type_comment=None,
        ),
    ],
    type_ignores=[],
)
```

```
$ cat t2.py 
x = 1  # type: int  # noqa
$ astpretty --typed-27 t2.py
Module(
    body=[
        Assign(
            lineno=1,
            col_offset=0,
            targets=[Name(lineno=1, col_offset=0, id='x', ctx=Store())],
            value=Num(lineno=1, col_offset=4, n=1),
            type_comment='int  # noqa',
        ),
    ],
    type_ignores=[],
)
```

sigmavirus24: Sorry, you should look a bit more closely at the second AST you pretty-printed, namely:

```
            type_comment='int  # noqa',
```

In other words, typed AST doesn't handle any case where some other linter's comment-based directives are present.

asottile: it works, the inner comments get discarded since they're re-parsed by the ast:

```
$ cat t.py
from typing import List
x = []  # type: List[int]  # noqa
reveal_type(x)
$ mypy t.py
t.py:3: error: Revealed type is 'builtins.list[builtins.int]'
```

asottile:
I rebased to resolve the conflict with #398

pyflakes/checker.py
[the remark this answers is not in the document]
asottile: yeah I thought about warnings, but there's no precedent in pyflakes currently.  basically I'd do

```python
if not tokens:
    warnings.warn(FutureWarning, 'tokens was not specified, not checking `# type:` comments')
```

though this might be annoying for those that can't control their environments

asottile:
@PeterJCLaw did you get a chance to try this on larger codebases?  I tried it against a few of our larger codebases at lyft with success

PeterJCLaw:
@asottile yup, just done that. Looks good (even found a few places where we had redundant imports :)

pyflakes/checker.py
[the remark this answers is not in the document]
PeterJCLaw: Ah, good point about this ending up showing for end users rather more than the developers of the intermediate packages.
I'm not sure what's best in that case. Perhaps just something in the docs?

tox.ini
```diff
@@ -6,11 +6,10 @@ envlist =
 [testenv]
 deps = flake8==3.6.0
 commands =
-    python setup.py test -q
+    python -m unittest discover pyflakes
+    flake8 --version
```
sigmavirus24: Is this left over from debugging?

asottile: seems it was an implicit conflict with https://github.com/PyCQA/pyflakes/pull/398/files#diff-b91f3d5bd63fcd17221b267e851608e8L10 from #398

asottile:
rebased and removed the `flake8 --version`

sigmavirus24:
This is awesome! Thanks @asottile!

asottile:
no problem! pretty excited about this change myself, it's really all that's stopping me from using mypy in 2+3 projects

sigmavirus24:
I'd love to start taking advantage of this in Flake8 proper.

asottile:
If pyflakes makes a release, I think all that's blocking flake8 3.7.0 would be the fixing of `per-file-ignores` 🎉

jriddy:
Do you guys know when you plan to cut a release with this in it?

Issue #404, opened 2019-01-02
Warn about `return` inside `finally`


joaoe:
Hi.

There is a feature in many programming languages (e.g. Java and python) in which a `return` from a `finally` block will silence the exception. It is a dangerous feature of the languages as errors can easily be suppressed unknowingly, e.g.
```
def task(data):
  try:
    complex_task(data)
  finally:
    if not data:
      return # in case of an exception, it is dropped
    logger.info("Blah...")
    another_complex_task(data)
    cleanup(data)
```
Silencing of exception should always be done explicitly with a proper `except ExceptionType` block with a `pass` or `return`. If you want to break from the `finally` block then you can refactor the code not to use a `return`.

Hence, I suggest this shall be flagged as a warning
```
def anything():
  raise ValueError()

def main():
  try:
    anything()
  finally:
    return

main()
```

joaoe:
Doh, someone incorrectly closed the pull request :(

asmeurer:
I didn't know about this behavior. It is [documented](https://docs.python.org/3/reference/compound_stmts.html#try), but the documentation also seems to imply that it is only done this way for technical reasons ("The exception information is not available to the program during execution of the finally clause."). 

I wonder if you can point to some references that this is universally bad behavior. pyflakes only warns about things that are 100% always wrong, to avoid false positives. It seems that doing a return in a finally could have a legitimate use-case. Granted, most people won't know that this happens, but I don't necessarily see that as a valid argument to disallow it from a linter (most people also don't know how for/else works, but that doesn't mean that it's a construct that should never be used). 

The closest parallel I can think of to this is a bare `except:` block. It's almost always used incorrectly, silencing more exceptions than desired, and it is not technically needed (it's syntactically equivalent to `except BaseException:`). But it does have legitimate uses, and if pyflakes warned about it (it doesn't), there would be a lot of false positives.

joaoe:
> The exception information is not available to the program during execution of the finally clause

That is not true:
```
import sys
try:
  raise ValueError()
finally:
  print(sys.exc_info())
```

> It seems that doing a return in a finally could have a legitimate use-case.
The two cases are: 
1. supress an exception, but it is the best practice to do that explicitly with an `except`  block
2. return from the function but the code can easily be refactored not to rely on the `return` inside the `finally`

Another broken behavior of `return` inside finally is when running `finally` after a return
```
def a():
	try:
		return 1
	finally:
		return 2

print(a())
```
[...]

Issue #403, opened 2019-01-02
Flag `return` inside generator function


joaoe:
It has happened often with me, having some big function (more than a couple screens length) that I change from a standard `return` logic to yielding a generator; or editing some generator function and putting a `return value` by accident, instead of `yield value`.

Python allows `return value` inside generator functions, which is equivalent to `raise StopIteration(value)`. However, I regard this an obscure feature of the language because `StopIteration` is not an error we should be catching but instead is transparently handled during for-loops. Hence, that `value` is never read.

So, in my opinion, a `return value` error in a generator is most likely a programming error which can be quite tricky to find unless you aim for 100% test coverage, to catch that pesky if-return check that handles some uncommon error.

I suggest that the following code is flagged as a warning:
```
anything = None # any value, really

def gen_function1():
  yield 42
  return anything

def gen_function2(weird_error):
  if weird_error:
    return anything
  yield 42
```
while the following is still allowed
```
def gen_function1():
  yield 42
  return

def gen_function2(weird_error):
  if weird_error:
    return
  yield 42
```

For the sake of backwards compatibility, and because I expect some people might disagree or have legitimate uses for `return value` inside the generator, this warning should be disableable, given there is no alternative way to do `raise StopIterator(value)` as the latter will trigger a `RuntimeError`.

Discuss :)

PS: I found this https://github.com/PyCQA/pyflakes/issues/166 so this is about enabling the warning always.

asottile:
this puzzlingly seems to only be enabled for python2.x

https://github.com/PyCQA/pyflakes/blob/009f4e7c21890cefafe5858aa682a7e7f17d6a80/pyflakes/checker.py#L1364-L1373

digging a little more, the condition was added in b24d50b88e26990118ef17d67ce6714c5e2ea31f

ah yes, for the coroutine things:

```python
def f(y):
    x = yield y
    return x * 9


g = f(3)
print(next(g))
try:
    g.send(2)
except StopIteration as e:
    print(e.args[0])
```

```console
$ python3 t.py
3
18
```

In python2.x it's a syntax error, python3 not:

```
$ python2 t.py 
  File "t.py", line 3
    return x * 9
SyntaxError: 'return' with argument inside generator
```

asottile:
pyflakes doesn't have a "disableable" aspect so I think this isn't actionable

joaoe:
> pyflakes doesn't have a "disableable" aspect so I think this isn't actionable

[...]

PR #405, opened 2019-01-02
Warn about return inside finally blocks

joaoe:
Issue #404

Returns inside finally blocks will suppress exceptions. Exceptions should be
suppressed explicitly in their except blocks. To break the flow in a finally block, refactor
it not to rely on the return.


sigmavirus24:
Based on the discussion on #404 I'm closing this since we do the right thing for Python 2.7

joaoe:
Hi. There was no discussion in #404.  You are confusing it with #403

joaoe:
Could you please reopen this ?

PR #406, opened 2019-01-04
Move type annotations tests to their own class

### LINKED ISSUE #402: TestAsync tests things that are not async, split up the test case
followup from https://github.com/PyCQA/pyflakes/pull/400#discussion_r244647478

asottile:
Resolves #402


pyflakes/test/test_type_annotations.py
```diff
@@ -0,0 +1,264 @@
...
+
+        def f(s):
+            return s
+
+        @typing.overload
+        def g(s):  # type: (None) -> None
+            pass
+
+        @typing.overload
+        def g(s):  # type: (int) -> int
```
sigmavirus24: Could we have an example that's like:

```py
@typing.overload
def g(s):
    # type: (int) -> int
    """Docstring."""
    pass
```

(or whatever the correct order is?

asottile: done!

PR #408, opened 2019-01-14
Add check for use of 'is' operator with str, bytes, and int literals

jdufresne:
While it often appears to "work", it is considered bad practice to compare against string, bytes, and int literals using the `is` operator. This is because it relies on a CPython implementation detail that may
not hold true for other interpreters. For example:

```py
    # Bad
    if x is 'foo':
        ...
```

```py
    # Good
    if x == 'foo':
        ...
```

This can easily be caught during static analysis to promote best practices. This is discussed in the Python bug: https://bugs.python.org/issue34850


pyflakes/checker.py
```diff
@@ -1662,3 +1662,15 @@ def ANNASSIGN(self, node):
...
             self.handleNode(node.value, node)
+
+    def COMPARE(self, node):
+        literals = (ast.Str, ast.Num)
+        if not PY2:
+            literals += (ast.Bytes,)
+        for op in node.ops:
+            if isinstance(op, (ast.Is, ast.IsNot)):
+                if (isinstance(node.left, literals) or
+                        any(isinstance(right, literals) for right in node.comparators)):
```
asottile: sorry for the drive by, this produces a false positive (however nonsensical):

```python
# evaluated as x is True and True < 4
x is True < 4
```

there's probably a less-nonsensical false-positive with `enum`s but gotta run to dinner

asottile:
overall like the idea though 👍

jdufresne:
Thanks for identifying that case! I have fixed it and added some tests for it. If you can demonstrate other false positives using enum, let me know and I'll add test for those as well.

It isn't clear to me why nightly is failing. Do you know if `ast.parse()` now parses `True`/`False` as `Num` instead of `Name`?

asottile:
It parses it as `Constant`:

```
$ ./python 
Python 3.8.0a0 (heads/master:995d9b9297, Jan 12 2019, 20:38:24) 
[GCC 7.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import ast
>>> ast.dump(ast.parse('if x is True: pass'))
"Module(body=[If(test=Compare(left=Name(id='x', ctx=Load()), ops=[Is()], comparators=[Constant(value=True)]), body=[Pass()], orelse=[])])"
```

asottile:
your guess appears to be correct, this is probably a regression:

```
(Pdb) left
<_ast.Name object at 0x7f8f2a99a370>
(Pdb) isinstance(left, literals)
False
(Pdb) n
> /tmp/pyflakes/pyflakes/checker.py(1674)COMPARE()
-> if isinstance(op, (ast.Is, ast.IsNot)):
(Pdb) n
> /tmp/pyflakes/pyflakes/checker.py(1675)COMPARE()
-> if isinstance(left, literals) or isinstance(right, literals):
(Pdb) p op
<_ast.IsNot object at 0x7f8f2b732a00>
(Pdb) p right
<_ast.Constant object at 0x7f8f2a99a2e0>
(Pdb) isinstance(right, literals)
True
(Pdb) p right
<_ast.Constant object at 0x7f8f2a99a2e0>
(Pdb) ast.dump(right)
'Constant(value=True)'
(Pdb) isinstance(right, ast.Num)
True
```

asottile:
I don't think you should work around the py38 behaviour -- pretty sure it's a bug: https://bugs.python.org/issue35733

can probably just set travis-ci to temporarily allow failures in py38

jdufresne:
Thanks for the help. I've updated the PR so now, in addition to fixing the previous false positives, the new tests pass on nightly. :slightly_smiling_face: 

It looks like some tests are failing on nightly, but they also fail on master for me, so I believe the failures are unrelated.

asottile:
I made a PR to cpython to fix the `Constant` regression -- https://github.com/python/cpython/pull/11547

(cpython is making some optimizations around parsing which is why there's a new `ast.Constant` type and the oddness with `isinstance`)

asottile:
fwiw, here's how to mark a matrix entry as "allowed to fail": https://docs.travis-ci.com/user/customizing-the-build/#rows-that-are-allowed-to-fail

I think for us that would just be adding:

```diff
     - python: nightly
       dist: xenial
+  allow_failures:
+    - python: nightly
+      dist: xenial
```

jdufresne:
Ah, thanks for the link to the Python bug. 

I've reverted the Python 3.8 workarounds and instructed Travis to ignore nightly failure. Thanks!

asottile:
and when https://github.com/PyCQA/pyflakes/pull/411 merges I think we can un-ignore :)

sigmavirus24:
Please rebase and unignore since #411 is merged

jdufresne:
I have rebased but Travis is still failing on nightly. Even though the CPython PR has been merged, I don't think it has made its way to Travis yet.

asottile:
yeah I was trying to find their documentation for how often they refresh nightly, the best I could find was "a recent version" which isn't exactly helpful.  I have my doubts that it's truly "nightly" but maybe we'll see if we try tomorrow!

sigmavirus24:
I re-ran nightly a litle bit ago and it passed so nightly seems to truly be nightly.

PR #415, opened 2019-01-21
Fix regression with `__all__ = [not, strings]`

asottile:
Regression introduced in 47a164136982ea6ef80f302200b2a29d3a4b1c3c

Example trace before fix:

```
...

  File "/home/asottile/workspace/pyflakes/pyflakes/checker.py", line 998, in handleNodeStore
    binding = ExportBinding(name, node.parent, self.scope)
  File "/home/asottile/workspace/pyflakes/pyflakes/checker.py", line 426, in __init__
    self.names += ast.literal_eval(source.value)
  File "/usr/lib/python3.6/ast.py", line 85, in literal_eval
    return _convert(node_or_string)
  File "/usr/lib/python3.6/ast.py", line 61, in _convert
    return list(map(_convert, node.elts))
  File "/usr/lib/python3.6/ast.py", line 84, in _convert
    raise ValueError('malformed node or string: ' + repr(node))
ValueError: malformed node or string: <_ast.Attribute object at 0x7f6d2bf37b70>
```

Found via [this file](https://github.com/pytest-dev/pytest-bdd/blob/6db3496ad2cd6434b6cc2cbaadf905fba4b9c215/pytest_bdd/__init__.py#L8)

CC @CLiu13

### DIFF
--- a/pyflakes/test/test_imports.py
+++ b/pyflakes/test/test_imports.py
@@ -1093,6 +1093,33 @@ class TestSpecialAll(TestCase):
         __all__ = ['a'] + ['b'] + ['c']
         ''', m.UndefinedExport, m.UndefinedExport, m.UndefinedExport, m.UnusedImport)
 
+    def test_all_with_attributes(self):
+        self.flakes('''
+        from foo import bar
+        __all__ = [bar.__name__]
+        ''')
+
+    def test_all_with_names(self):
+        # not actually valid, but shouldn't produce a crash
+        self.flakes('''
+        from foo import bar
+        __all__ = [bar]
+        ''')
+
+    def test_all_with_attributes_added(self):
+        self.flakes('''
+        from foo import bar
+        from bar import baz
+        __all__ = [bar.__name__] + [baz.__name__]
+        ''')
+
+    def test_all_mixed_attributes_and_strings(self):
+        self.flakes('''
+        from foo import bar
+        from foo import baz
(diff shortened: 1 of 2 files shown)


asottile:
actually, this still has problems now that I look at the other patch -- let me try and restore the original behaviour plus the new behaviour

asottile:
ok, should be good now -- no more literal_eval

pyflakes/test/test_imports.py
```diff
@@ -1093,6 +1093,26 @@ def test_concatenationAssignment(self):
...
+        from bar import baz
+        __all__ = [bar.__name__] + [baz.__name__]
+        ''')
+
+    def test_all_mixed_attributes_and_strings(self):
+        self.flakes('''
+        from foo import bar
+        from foo import baz
+        __all__ = ['bar', baz.__name__]
+        ''')
```
sigmavirus24: I've also (in the past) written (and continue to see others do)

```py
from foo import bar
from foo import baz

__all__ = [bar, baz]
```

Should we test for that too? I'm not sure we should necessarily do the right thing but I do wonder if we want to test the behaviour at least.

asottile: I'll add a test that makes sure it doesn't crash.  That said, afaict it's not valid:

```python
import os

__all__ = [os]
```

```pycon
>>> from t import *
TypeError: attribute name must be string, not 'module'
```

PR #416, opened 2019-01-21
Allow star args in type comments

asottile:
Found while trying out the release on [this file](https://github.com/lyft/awspricing/blob/23493d2c397da9dcffcf3b598df5b5d32f303121/awspricing/constants.py#L14)

### DIFF
--- a/pyflakes/test/test_type_annotations.py
+++ b/pyflakes/test/test_type_annotations.py
@@ -231,6 +231,14 @@ class TestTypeAnnotations(TestCase):
             return a + b
         """)
 
+    def test_typeCommentsStarArgs(self):
+        self.flakes("""
+        from mod import A, B, C, D
+        def f(a, *b, **c):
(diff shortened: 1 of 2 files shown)


pyflakes/test/test_type_annotations.py
```diff
@@ -231,6 +231,14 @@ def f(a, b):
             return a + b
         """)
 
+    def test_typeCommentsStarArgs(self):
+        self.flakes("""
+        from mod import A, B, C, D
+        def f(a, *b, **c):
+            # type: (A, *B, **C) -> D
```
sigmavirus24: That's a valid type sig? I didn't know that.

asottile: most of the time I've seen it `C` will end up being `Any`

as in:
```python
def f(*args: str, **kwargs: Any) -> None:
   ...
```

Issue #425, opened 2019-02-02
Include colon number in the diagnostics


AmjadHD:
pyflakes only provides the line number for diagnostics, It would be very handy to include the colon number as well for more precise diagnostics.
related issues: #417 #158 .

PR #426, opened 2019-02-07
Include column numbers with all error messages

asmeurer:
As long as the node passed to the message has a col_offset, it is included in
the message. There is already a precedent of printing the column number for
syntax errors.

This changes the API slightly, in that if col_offset is not present on the
node, the message.col is now -1 instead of 0. Otherwise, there would be no way
to distinguish between errors that have no column number and errors with a
column number 0. If this doesn't seem like an acceptable change, I can revert
it, and either always use 0 as the default, or use a separate attribute to
indicate if the message.col is the "real" column number.

I haven't yet tested everything, but I'm not even sure if there are any nodes
that don't have a col_offset. There don't appear to be any represented in the
test suite.

I haven't added any tests yet. I want to get feedback first if this is a desired change. 

Fixes https://github.com/PyCQA/pyflakes/issues/158, fixes https://github.com/PyCQA/pyflakes/issues/425.

### DIFF
--- a/pyflakes/messages.py
+++ b/pyflakes/messages.py
@@ -10,10 +10,10 @@ class Message(object):
     def __init__(self, filename, loc):
         self.filename = filename
         self.lineno = loc.lineno
-        self.col = getattr(loc, 'col_offset', -1)
+        self.col = getattr(loc, 'col_offset', -1) + 1
 
     def __str__(self):
-        column = '%s:' % self.col if self.col != -1 else ''
+        column = '%s:' % self.col if self.col else ''
         return '%s:%s:%s %s' % (self.filename, self.lineno, column,
                                  self.message % self.message_args)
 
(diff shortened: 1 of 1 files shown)


asmeurer:
Also found this related issue https://github.com/PyCQA/pyflakes/issues/425. 

And https://github.com/PyCQA/pyflakes/issues/169 points to #15 as adding the column number for syntax errors, but that is obviously not the right PR. Did the PR numbers change at some point?

asmeurer:
OK, I've pushed what I think is the best compromise here, which minimizes API breakage. See the commit message on my latest commit. The only oddity here is that if a node without a col_offset (like a Module node) led to a message, it would print a column number of 1. But I don't think this is a big deal as 1 is a perfectly reasonable default value, and, more importantly, no such nodes are currently do this, at least none represented in the tests (no tests fail if I replace `getattr(loc, 'col_offset', 0)` with `loc.col_offset`).

Issue #422, opened 2019-01-30
2.1.0 is broken


ThomasWaldmann:
https://travis-ci.org/borgbackup/borg/jobs/486513787

asottile:
while this isn't a very quality bug report it gave me enough information to find a minimal reproduction.  @ThomasWaldmann  in the future it would be good to include the stacktrace, the project, the version information, etc. in the issue (even the filename if you could identify it)

I've managed to boil this down to a minimal case:

```python
x = lambda f: f

@x
def t():
    pass

y = lambda f: f

@x
@y
def t():
    pass

@x
@y
def t():
    pass
```

asottile:
#423 has a fix for this

sigmavirus24:
That's a question for @bitglue, ~grumpy cat~ @ThomasWaldmann

ThomasWaldmann:
Update: besides failing all our travis-ci test runs, it also hangs flake8 in our local vagrant-based test setup, see link directly above.

sfstpala:
This issue is not fixed

sfstpala:
Deleted my comment. I thought you guys would be more welcoming. 👎

PR #433, opened 2019-02-21
Add another test for #422

untitaker:
#422


asottile:
given there's already a test which covers this regression -- is there benefit for two?

untitaker:
idk if this goes through a different codepath or not, feel free to close

asottile:
this is definitely a simpler case of the one I tried to isolate -- in the end it's the same code path though

untitaker:
@asottile do you want me to replace your test or close this?

asottile:
replace is probably good 👍

PR #435, opened 2019-02-22
Improve @overload detection

### LINKED ISSUE #434: F811 (redefinition of unused variable) incorrectly triggered with @overload decorator
This bug was fixed for functions in #320. It still occurs if the function is inside a class and/or has multiple decorators (e.g. @staticmethod, @classmethod):
```python
from typing import overload


class test:
    @overload
    def utf8(value: None) -> None:
        pass

    @overload
    def utf8(value: bytes) -> bytes:
        pass

    @overload
    def utf8(value: str) -> bytes:
        pass

    def utf8(value):
        pass  # actual implementation
```

```
test.py:9:5: F811 redefinition of unused 'utf8' from line 5
test.py:13:5: F811 redefinition of unused 'utf8' from line 9
test.py:17:5: F811 redefinition of unused 'utf8' from line 13
```

```python
from typing import overload


def do_nothing(func):
    return func


@overload
@do_nothing
def utf8(value: None) -> None:
    pass


@overload
@do_nothing
def utf8(value: bytes) -> bytes:
    pass


@overload
@do_nothing
def utf8(value: str) -> bytes:
    pass


def utf8(value):
    pass  # actual implementation
```

```
test.py:14:1: F811 redefinition of unused 'utf8' from line 8
test.py:20:1: F811 redefinition of unused 'utf8' from line 14
test.py:26:1: F811 redefinition of unused 'utf8' from line 20
```

asottile:
Resolves #434

### DIFF
--- a/pyflakes/checker.py
+++ b/pyflakes/checker.py
@@ -548,8 +548,10 @@ def is_typing_overload(value, scope):
 
     return (
         isinstance(value.source, ast.FunctionDef) and
-        len(value.source.decorator_list) == 1 and
-        is_typing_overload_decorator(value.source.decorator_list[0])
+        any(
+            is_typing_overload_decorator(dec)
+            for dec in value.source.decorator_list
+        )
     )
 
 
--- a/pyflakes/checker.py
+++ b/pyflakes/checker.py
@@ -530,14 +530,21 @@ def getNodeName(node):
         return node.name
 
 
-def is_typing_overload(value, scope):
+def is_typing_overload(value, scope_stack):
+    def name_is_typing_overload(name):  # type: (str) -> bool
+        for scope in reversed(scope_stack):
+            if name in scope:
+                return (
+                    isinstance(scope[name], ImportationFrom) and
+                    scope[name].fullName == 'typing.overload'
+                )
+        else:
+            return False
+
     def is_typing_overload_decorator(node):
         return (
             (
-                isinstance(node, ast.Name) and
-                node.id in scope and
-                isinstance(scope[node.id], ImportationFrom) and
-                scope[node.id].fullName == 'typing.overload'
+                isinstance(node, ast.Name) and name_is_typing_overload(node.id)
             ) or (
                 isinstance(node, ast.Attribute) and
                 isinstance(node.value, ast.Name) and
@@ -890,7 +897,7 @@ class Checker(object):
                                 node, value.name, existing.source)
                 elif not existing.used and value.redefines(existing):
                     if value.name != '_' or isinstance(existing, Importation):
-                        if not is_typing_overload(existing, self.scope):
+                        if not is_typing_overload(existing, self.scopeStack):
                             self.report(messages.RedefinedWhileUnused,
                                         node, value.name, existing.source)
 
--- a/pyflakes/test/test_type_annotations.py
+++ b/pyflakes/test/test_type_annotations.py
@@ -39,6 +39,25 @@ class TestTypeAnnotations(TestCase):
             return s
         """)
 
+    def test_overload_with_multiple_decorators(self):
+        self.flakes("""
+            from typing import overload
+            dec = lambda f: f
+
+            @dec
+            @overload
+            def f(x):  # type: (int) -> int
+                pass
+
+            @dec
+            @overload
+            def f(x):  # type: (str) -> str
+                pass
+
+            @dec
+            def f(x): return x
+       """)
+
     def test_not_a_typing_overload(self):
         """regression test for @typing.overload detection bug in 2.1.0"""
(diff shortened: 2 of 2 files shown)


Zac-HD:
Hurray!  

This fixes another bug too: in the current released version, using a decorator named `overload` causes an internal error (because pyflakes thinks that it's `typing.overload`, and tries to access a nonexistent scope name).  We first noticed this a while ago in [Hypothesis' build](https://travis-ci.org/HypothesisWorks/hypothesis/jobs/498740860#L783), but it's reproducible by running `pyflakes` on the following file:

```python
overload = lambda f: f  # or `def overload(arg): return arg

@overload
def func(arg): pass

def func(arg): return ...
```

Happily I've confirmed that this change fixes pyflakes for both the minimal case above (which might be worth adding as another test case for this PR?) and for Hypothesis - thanks @asottile 😍

asottile:
@Zac-HD 🎉 though it was probably #423 that fixed that one ~o/

Zac-HD:
Yeah, probably... still fixed though and still fixed by you 😁

asottile:
heh only fair -- I broke it in the first place /o\

kelseyfrancis:
Thanks for fixing this! Could you all please cut a pypi release that includes this fix?

omry:
When is this expected to make it to flake8?

asottile:
no current plan, but you can read my thoughts [here](https://github.com/PyCQA/pyflakes/issues/494#issuecomment-561303887) or as a temporary workaround install from `git+https://github.com/pycqa/pyflakes@revision_here`

omry:
Looks like this issue still happens with master on python 3.8.

asottile:
doesn't happen for me?  try uninstalling `pyflakes` before installing from `git` (I suspect you're not using the actual git install)

omry:
You are right, I had a system wide flake8 installed and bash already had it cached as flake8.

omry:
@ asottile, unfortunately depending on a git version in setup.py prevents publishing a package.
would you be able to make a dev/pre release?

asottile:
it's a bit unusual to depend on pyflakes in `setup.py` -- could you elaborate on the usecase?

(I also don't have pypi permission, I believe @bitglue manages that)

omry:
Sure:
I am depending on flake for linting my code [here](https://github.com/omry/omegaconf/blob/11dd822e5ca0265216896cb6ccd5c7d451544146/setup.py).

Because of this overload bug, I am also depending on pyflake to get this bugfix.
This dependency (as well as the one with isort which is due to a similar problem) are preventing me from publishing the package as is (and I need to hack those out before publishing).

asottile:
ah I see -- those dependencies aren't for your consumers so there isn't really a need to put them in `setup.py` -- maybe put them in your `nox` file instead?

omry:
developers are also consumers.
I am using "pip install -e .[dev]" to setup the development environment.

my noxfile is also [settings up](https://github.com/omry/omegaconf/blob/11dd822e5ca0265216896cb6ccd5c7d451544146/noxfile.py#L48) in the same way.

asottile:
yeah I notice you're using pre-commit -- I'm the author (you've got it set up in a less-than-supported manner -- first you don't need `.pre-commit-hooks.yaml` that's for _providing_ hooks, next you're using the escape hatch `local` + `system` hooks instead of versioning directly in `.pre-commit-config.yaml`) -- but that's a separate discussion entirely :)

when I said consumers above, I meant people that would install your package -- I understand you still want to have a developer workflow, I'm just suggesting that while `setup.py` does work for that, it isn't really the intended mechanism (whereas something more for your developer workflow such as `nox` / `tox` / `poetry` or even `requirements-dev.txt` is more properly suited for that)

omry:
Thanks for the pointers! didn't realize you are the author of pre-commit hooks.
About that, I was following the instructions from Black, I was a bit confused about the purpose of the second file.
Looking now, looks like Black got [better instructions](https://black.readthedocs.io/en/stable/version_control_integration.html). I update and clean up the mess.

omry:
I [fixed](https://github.com/omry/omegaconf/blob/87d903d80d4cac6ae209d26f18004f99833b0912/.pre-commit-config.yaml) the pre-commit config for black to not use the local escape hatch, but I am not sure how to do it for flake8 given the need to install a specific revision of pyflakes.

asottile:
something like this:

```yaml
-   repo: https://gitlab.com/pycqa/flake8
    rev: 3.7.9
    hooks:
    -   id: flake8
        additional_dependencies: [-e, git+https://github.com/pycqa/pyflakes@...]
```

omry:
Awesome, thanks!

PR #438, opened 2019-03-02
Shebang detection improvements

scop:


bitglue:
Would you please add a test for the case where there's an otherwise valid shebang, but it's not on the first line?

PR #439, opened 2019-03-02
Remove unnecessary `else`

myint:
This merely simplifies the code by removing an unnecessary `else`.

### DIFF
--- a/pyflakes/checker.py
+++ b/pyflakes/checker.py
@@ -538,8 +538,8 @@ def is_typing_overload(value, scope_stack):
                     isinstance(scope[name], ImportationFrom) and
                     scope[name].fullName == 'typing.overload'
                 )
-        else:
-            return False
+
+        return False
 
     def is_typing_overload_decorator(node):
         return (


asottile:
fwiw, I put it there because it's easier for humans to read and doesn't affect the execution (it's much clearer that the for loop either early returns, or when it doesn't it returns False)

myint:
I'm mostly just patching this as a reaction to static analysis:

https://lgtm.com/projects/g/PyCQA/pyflakes/alerts/?mode=list

Also, I had to remind myself again what `else` does here.

sigmavirus24:
I'm on the fence, personally. I agree with @asottile that this makes it clearer about the logic, but I also know that probably ~80% of Python developers don't understand or remember how `for/else` works and this code is semi-clearer for the majority without the `else`. Should we rob people of a chance to learn something? :man_shrugging: 

I think the tool (LGTM, or whatever it is using) has a point though. `for/else` makes the most sense when you want to catch when a `for` loop didn't exit before consuming what it's iterating over. With a return in the for-loop body, it should be clear enough that you will return False if you didn't return something else first.

I think that means I've talked myself into merging this.

PR #443, opened 2019-04-04
String formatting linting

### LINKED ISSUE #148: Detect broken string templates that will raise ValueError when interpolated

asottile:
There's three classes of lint messages added here -- I'm happy to split out / reorder / etc. these commits as fit.   Also please suggest better names / messages -- these were my first-pass attempt at the messages 😆 

Resolves #148 

Supersedes #370 (CC @fperrin)

### f-strings

- f-strings without placeholders

### `'...'.format(...)`

- `'{}'.format(1, 2)` -- extra positional arguments
- `'{foo}'.format(foo=1, bar=2)` -- extra named arguments
- `'{} {}'.format(1)` -- placeholder missing argument (positional)
- `'{foo} {bar}'.format(foo=1)` -- placeholder missing argument (named)
- `'{} {1}'.format(1, 2)` -- switching between manual / automatic numbering
- `'{0} {}'.format(1, 2)` -- switching between manual / automatic numbering
- `'{'.format(1)` -- invalid format string

### `'...' % ...`

- `'%(foo)' % ()` -- invalid format string (missing the conversion type)
- `'%j' % ()` -- invalid format specifier (`j` is not a valid conversion type)
- `'%s %s' % (1, 2, 3)` -- positional count mismatch (too many args)
- `'%s %s' % (1,)` -- positional count mismatch (too few args)
- `'%(bar)s' % {'bar': 1, 'baz': 2}` -- extra named argument supplied
- `'%(bar)s %(baz)s' % {'bar': 1}` -- missing named argument
- `'%(bar)s' % (1, 2, 3)` -- expected mapping, but got sequence
- `'%s %s' % {'baz': 1}` -- expected sequence, got mapping

---

- Code for parsing `.format()` uses [`string.Formatter.parse`](https://docs.python.org/3/library/string.html#string.Formatter.parse)
- Code for parsing `%` format is lifted from [pyupgrade](https://github.com/asottile/pyupgrade): [here](https://github.com/asottile/pyupgrade/blob/cf4555014f9332ac06aebbca9dccfe75090be3d7/pyupgrade.py#L521-L578)

### DIFF
--- a/pyflakes/messages.py
+++ b/pyflakes/messages.py
@@ -306,3 +306,59 @@ class StringDotFormatInvalidFormat(Message):
     def __init__(self, filename, loc, error):
         Message.__init__(self, filename, loc)
         self.message_args = (error,)
+
+
+class PercentFormatInvalidFormat(Message):
+    message = "'...' %% ... has invalid format string: %s"
+
+    def __init__(self, filename, loc, error):
+        Message.__init__(self, filename, loc)
+        self.message_args = (error,)
+
+
+class PercentFormatMixedPositionalAndNamed(Message):
+    message = "'...' %% ... has mixed positional and named placeholders"
+
+
+class PercentFormatUnsupportedFormatCharacter(Message):
+    message = "'...' %% ... has unsupported format character %r"
+
+    def __init__(self, filename, loc, c):
+        Message.__init__(self, filename, loc)
+        self.message_args = (c,)
+
+
+class PercentFormatPositionalCountMismatch(Message):
+    message = "'...' %% ... has %d placeholder(s) but %d substitution(s)"
+
+    def __init__(self, filename, loc, n_placeholders, n_substitutions):
+        Message.__init__(self, filename, loc)
+        self.message_args = (n_placeholders, n_substitutions)
+
+
+class PercentFormatExtraNamedArguments(Message):
+    message = "'...' %% ... has unused named argument(s): %s"
+
+    def __init__(self, filename, loc, extra_keywords):
+        Message.__init__(self, filename, loc)
+        self.message_args = (extra_keywords,)
+
+
+class PercentFormatMissingArgument(Message):
+    message = "'...' %% ... is missing argument(s) for placeholder(s): %s"
+
+    def __init__(self, filename, loc, missing_arguments):
+        Message.__init__(self, filename, loc)
+        self.message_args = (missing_arguments,)
+
+
+class PercentFormatExpectedMapping(Message):
+    message = "'...' %% ... expected mapping but got sequence"
+
+
+class PercentFormatExpectedSequence(Message):
+    message = "'...' %% ... expected sequence but got mapping"
+
+
+class PercentFormatStarRequiresSequence(Message):
+    message = "'...' %% ... `*` specifier requires sequence"
--- a/pyflakes/messages.py
+++ b/pyflakes/messages.py
@@ -266,3 +266,7 @@ class InvalidPrintSyntax(Message):
 
 class IsLiteral(Message):
     message = 'use ==/!= to compare str, bytes, and int literals'
+
+
+class FStringMissingPlaceholders(Message):
+    message = 'f-string is missing placeholders'
--- a/pyflakes/messages.py
+++ b/pyflakes/messages.py
@@ -270,3 +270,39 @@ class IsLiteral(Message):
 
 class FStringMissingPlaceholders(Message):
     message = 'f-string is missing placeholders'
+
+
+class StringDotFormatExtraPositionalArguments(Message):
+    message = "'...'.format(...) has unused arguments at position(s): %s"
+
+    def __init__(self, filename, loc, extra_positions):
+        Message.__init__(self, filename, loc)
+        self.message_args = (extra_positions,)
+
+
+class StringDotFormatExtraNamedArguments(Message):
+    message = "'...'.format(...) has unused named argument(s): %s"
+
+    def __init__(self, filename, loc, extra_keywords):
+        Message.__init__(self, filename, loc)
+        self.message_args = (extra_keywords,)
+
+
+class StringDotFormatMissingArgument(Message):
+    message = "'...'.format(...) is missing argument(s) for placeholder(s): %s"
+
+    def __init__(self, filename, loc, missing_arguments):
+        Message.__init__(self, filename, loc)
+        self.message_args = (missing_arguments,)
+
+
+class StringDotFormatMixingAutomatic(Message):
+    message = "'...'.format(...) mixes automatic and manual numbering"
+
+
+class StringDotFormatInvalidFormat(Message):
+    message = "'...'.format(...) has invalid format string: %s"
+
+    def __init__(self, filename, loc, error):
+        Message.__init__(self, filename, loc)
+        self.message_args = (error,)
--- a/pyflakes/test/test_other.py
+++ b/pyflakes/test/test_other.py
@@ -1789,6 +1789,60 @@ class TestStringFormatting(TestCase):
             print(f'{x:>2} {y:>2}')
         ''')
 
+    def test_invalid_dot_format_calls(self):
+        self.flakes('''
+            '{'.format(1)
+        ''', m.StringDotFormatInvalidFormat)
+        self.flakes('''
+            '{} {1}'.format(1, 2)
+        ''', m.StringDotFormatMixingAutomatic)
+        self.flakes('''
+            '{0} {}'.format(1, 2)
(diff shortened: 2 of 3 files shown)


asmeurer:
Cool that it doesn't actually evaluate the string so it isn't susceptible to https://github.com/PyCQA/pyflakes/pull/370#issuecomment-427460642. Maybe that should be added as a test.

asottile:
Some examples of things this found in my code:

- [`all_repos/autofix_lib.py:171:25 f-string is missing placeholders`](https://github.com/asottile/all-repos/blob/4b5fd1f39c2a6bf80b0b85b197ae5cf361bbb8b3/all_repos/autofix_lib.py#L171)
- [`go_module_no_pc/build_go_module.py:36:11 '...'.format(...) has unused named argument(s): file
`](https://github.com/asottile/go-python-module-example/blob/485046bc810a08f841eb5091cc1dac40a1193489/go_module_no_pc/build_go_module.py#L36-L42)

And other code I have access to at a glance:

- [`scripts/release.py:95:11 f-string is missing placeholders`](https://github.com/pytest-dev/pytest/blob/13a9d876f74f17907ad04b13132cbd4aa4ad5842/scripts/release.py#L95)
- [`paasta_itests/steps/cleanup_marathon_job_steps.py:39:15 '...'.format(...) has unused arguments at position(s): 2`](https://github.com/Yelp/paasta/blob/46b3c2e90f26810d149efcffddc7607a71b9e275/paasta_itests/steps/cleanup_marathon_job_steps.py#L39-L42)
- [`paasta_tools/chronos_serviceinit.py:288:9 '...' % ... has unused named argument(s): is_temporary`](https://github.com/Yelp/paasta/blob/46b3c2e90f26810d149efcffddc7607a71b9e275/paasta_tools/chronos_serviceinit.py#L288-L304)
- [`paasta_tools/paasta_serviceinit.py:166:21 '...'.format(...) has unused arguments at position(s): 0, 1`](https://github.com/Yelp/paasta/blob/46b3c2e90f26810d149efcffddc7607a71b9e275/paasta_tools/paasta_serviceinit.py#L165-L167)
- [`paasta_tools/cli/cmds/mark_for_deployment.py:751:34 f-string is missing placeholders
`](https://github.com/Yelp/paasta/blob/46b3c2e90f26810d149efcffddc7607a71b9e275/paasta_tools/cli/cmds/mark_for_deployment.py#L751)

asottile:
Oops this is still WIP -- found two more cases to handle:

```python
# currently flagged as "too many placeholders"
'%.*f %s' % (2, b / float(factor), suffix)
# currently flagged as "unused placeholder at position 2"
'{:{}} {}'.format(1, 15, 2)
```

asmeurer:
Is it possible to have real column numbers here? For example

```py
"abc %t" % 1
```

reports "unsupported format character 't'" line 1, column 1. It would be nice to have column 7 (where the `t` is).

asmeurer:
Does this have any support for finding multiple errors in the same string? It seems like it could be another advantage of this approach. I tried `"%s%t" % 1` but it only gave one error.

asottile:
> Is it possible to have real column numbers here? For example
> 
> ```python
> "abc %t" % 1
> ```
> 
> reports "unsupported format character 't'" line 1, column 1. It would be nice to have column 7 (where the `t` is).

it's difficult, but doable -- I'd like to leave that to a v2 as a later improvement because it is not easy and purely cosmetic.  pyflakes currently only has support for node-based reporting and it would be a large refactor to point at positions within a node: https://github.com/PyCQA/pyflakes/blob/232cb1d27ee134bf96adc8f37e53589dc259b159/pyflakes/messages.py#L12-L13

> Does this have any support for finding multiple errors in the same string? It seems like it could be another advantage of this approach. I tried `"%s%t" % 1` but it only gave one error.

It does, however there's only 1 statically determinable error in that format string you've posted (without attempting to solve an intractably large problem of full type checking) -- notably the positional / keyword can only really be determined if it's a `tuple` / `list` / `dict` literal (and for the case of dictionaries it can only determine it if all the keys are strings)

Here's an example where multiple are detected:

```console
$ python3 -m pyflakes t.py
t.py:1:1 '...' % ... has unused named argument(s): bar
t.py:1:1 '...' % ... is missing argument(s) for placeholder(s): foo
$ cat t.py
'%(foo)s' % {'bar': 'baz'}
```

asmeurer:
Ah. I had assumed it would take literals as obviously not unpackable. But I can see that in general pyflakes would need a recursive type inference system, which it currently does not have. I guess this just shows why `format` is better than `%`.

asottile:
It ~could probably inference int / float / str / bool / None _literals_ without too much more complexity -- though I left them out since they're not interesting formatting

asmeurer:
That's a good point.

asottile:
here's a few more cases it found: https://github.com/python/mypy/pull/6631

asottile:
Found one more unhandled case that I'll get a fix for:

- [ ]
  ```python
   '{foo}-{}'.format(1, foo=2)
  ```

asottile:
I've tried this out on quite a few internal codebases and it appears to be stable and has proved quite useful (fixed a bunch of bugs in rare error cases!) -- would definitely appreciate some reviews on the approach with the hope that this gets integrated :)

pyflakes/messages.py
```diff
@@ -266,3 +266,99 @@ class InvalidPrintSyntax(Message):
...
+class PercentFormatUnsupportedFormatCharacter(Message):
+    message = "'...' %% ... has unsupported format character %r"
+
+    def __init__(self, filename, loc, c):
+        Message.__init__(self, filename, loc)
+        self.message_args = (c,)
+
+
+class PercentFormatPositionalCountMismatch(Message):
+    message = "'...' %% ... has %d placeholders but %d substitutions"
```
asmeurer: This should either use `(s)` or add or remove the `s` conditionally if based on if the value is > 1.

asmeurer:
I ran this on a bunch of codebases. I didn't really find many errors. Only a handful of issues that wouldn't lead to runtime errors (extra unused formatting arguments). But there were no crashes or incorrect reports, so that's good. I have no idea if it missed anything.

Some of the error messages can be a bit confusing. The arguments to the `format` function call are counted starting at 0. I can see why this is done, as that's how formatting works, but for something like `f(x, y, z)` I'm used to calling `z` the 3rd argument, not position 2. Maybe instead of listing the position it could name the argument itself. 

It can also be a bit confusing with something like `'{} {}'.format(1)` that the `{}` is called "placeholder 1", same as `'{0} {1}'.format(1)`. I can't think of a cleaner way to write this error message, however. I'm also not clear from the code if it would be straightforward distinguish `'{} {}'` and `'{0} {1}'` (I haven't read the code especially closely, though). 

Another suggestion would be to replace the `(s)` in the error messages with an automatic pluralizer to improve readability. Alternately emit one message per error. So for instance

 ```py
'{a} {b} {c}'.format(a=1)
```

would produce

```
'...'.format(...) is missing an argument for placeholder 'b'
'...'.format(...) is missing an argument for placeholder 'c'
```

instead of

```
'...'.format(...) is missing argument(s) for placeholder(s): b, c
```

I think I would prefer splitting like this, as this matches the way pyflakes handles other errors, and it will allow better reporting once column numbers are added (the 'b' and 'c' errors could each point to their respective position in the format string).

In all I think the changes here are pretty good. But again I didn't read the code too closely, so you might want someone else to do that.

File: pyflakes/messages.py (hunk 1 above)
Comment (asottile) [reply pyflakes/messages.py]: Going with `(s)` for now -- good catch on this one!

File: pyflakes/test/test_other.py [hunk 2]
```diff
@@ -1755,6 +1755,156 @@ def test_f_string(self):
...
+            k = {}
+            "{foo}".format(**k)
+        ''')
+
+    def test_invalid_percent_format_calls(self):
+        self.flakes('''
+            '%(foo)' % {'foo': 'bar'}
+        ''', m.PercentFormatInvalidFormat)
+        self.flakes('''
+            '%s %(foo)s' % {'foo': 'bar'}
```
myint: To fully exercise the code for this check, it would be good to add a similar test, but with '%(foo)s %s'.

asottile: done! thanks for the suggestion

asmeurer:
Should we open a new issue for splitting out the errors into separate errors for each item and adding column information?

PR #455, opened 2019-07-19
In PEP 484 type comments, allow text after "# type: ignore"

msullivan:
This is to support allowing typecheckers to implement ignores for
specific errors, using syntax like `# type: ignore=E1000` or
`# type: ignore[type-mismatch]` or some such.
mypy is about to add support for ignoring specific errors following
this design: https://github.com/python/mypy/issues/7239

Support for extra text in type comments was implemented
in CPython as https://bugs.python.org/issue36878
and in typed_ast as https://github.com/python/typed_ast/pull/116.

### DIFF
--- a/pyflakes/test/test_checker.py
+++ b/pyflakes/test/test_checker.py
@@ -152,10 +152,6 @@ class CollectTypeCommentsTests(TestCase):
         ret = self._collect('x = 1 #type:int')
         self.assertSetEqual(ret, {(ast.Assign, ('#type:int',))})
 
-    def test_type_comment_starts_with_word_ignore(self):
-        ret = self._collect('x = 1 # type: ignore[T]')
-        self.assertSetEqual(ret, {(ast.Assign, ('# type: ignore[T]',))})
-
     def test_last_node_wins(self):
         """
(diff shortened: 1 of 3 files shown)


pyflakes/test/test_checker.py
```diff
@@ -152,10 +152,6 @@ def test_type_comment_without_whitespace(self):
         ret = self._collect('x = 1 #type:int')
         self.assertSetEqual(ret, {(ast.Assign, ('#type:int',))})
 
-    def test_type_comment_starts_with_word_ignore(self):
-        ret = self._collect('x = 1 # type: ignore[T]')
-        self.assertSetEqual(ret, {(ast.Assign, ('# type: ignore[T]',))})
```
asottile: so it's more clear of the change here -- can this be updated with the new behaviour?

msullivan: Done

wsanchez:
Any release planned so we can see this fix on PyPI?

asottile:
> Any release planned so we can see this fix on PyPI?

not yet, please be patient (this comment is not productive) we'll release when we're ready!

PR #467, opened 2019-09-04
Fix issue 428

### LINKED ISSUE #428: Pyflakes doesn't properly handle del inside function definition
```py
a = 1

def test():
    a = 2
    del a
    print(a)

test()
```

The code produces `UnboundLocalError: local variable 'a' referenced before assignment`, but pyflakes shows now errors. 

The reason this code raises an exception is that local variables are determined statically at compile time, not at runtime, so the `a = 2` line forces `a` to be function local inside of `test`. The `del a` does not make it global again (it is impossible to "make a local variable global again").

Hanaasagi:
It will resolve #428

### DIFF
--- a/pyflakes/checker.py
+++ b/pyflakes/checker.py
@@ -16,6 +16,11 @@ import string
 import sys
 import tokenize
 
+try:
+    from collections import UserDict
+except ImportError:
+    from UserDict import IterableUserDict as UserDict
+
 from pyflakes import messages
 
 PY2 = sys.version_info < (3, 0)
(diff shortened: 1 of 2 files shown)


Hanaasagi:
Know little about PyPy, it's behavior is different from Python :joy:

asottile:
looks like new versions of pypy have different offsets, I'll make a PR to fix that -- should be unrelated to your change here 👍

Hanaasagi:
Do you have any questions about this PR?

PR #472, opened 2019-09-23
Allow @overload on async functions as well

### LINKED ISSUE #470: @overload of async functions
Since #435, pyflakes git tip copes with synchronous @overload'ed functions, but not with async ones:

```python
from typing import overload

@overload
def f(x: int) -> int:
    ...

@overload
def f(x: float) -> float:
    ...

def f(x):
    return x

@overload
async def g(x: int) -> int:
    ...

@overload
async def g(x: float) -> float:
    ...

async def g(x):
    return x
```
```bash
$ python -m pyflakes .

18:1 redefinition of unused 'g' from line 14
22:1 redefinition of unused 'g' from line 18
```

asottile:
resolves #470

### DIFF
--- a/pyflakes/checker.py
+++ b/pyflakes/checker.py
@@ -72,9 +72,11 @@ else:
 if PY35_PLUS:
     FOR_TYPES = (ast.For, ast.AsyncFor)
     LOOP_TYPES = (ast.While, ast.For, ast.AsyncFor)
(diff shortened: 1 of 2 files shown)


crusaderky:
Tested manually; it works

asottile:
(counting that as two approvals :laughing:)

PR #476, opened 2019-10-01
Allow continue inside finally in 3.8+

isidentical:
See [bpo-32489](https://bugs.python.org/issue32489)


asmeurer:
Also apparently https://www.python.org/dev/peps/pep-0601/ was rejected.

asottile:
> Also apparently https://www.python.org/dev/peps/pep-0601/ was rejected.

yep, @isidentical is one of the authors \o/

PR #479, opened 2019-10-05
Fix false positive with partially quoted annotations

### LINKED ISSUE #447: Spurious import warning for type used in nested quoted annotation
This is similar to #290 and #313, but seems to still be an issue.

If you import a type and then use it only within a type annotation which includes a nested quoted part, then `pyflakes` reports the import as unused.

``` python
from queue import Queue  # F401: 'queue.Queue' imported but unused
from typing import Optional

def foo(queue: Optional['Queue[int]'] = None) -> None:
    print(queue)

# Revealed type is 'def (queue: Union[queue.Queue[builtins.int], None] =)'
reveal_type(foo)
```

This can be worked around by quoting the whole of the type annotation:
``` python
def foo(queue: 'Optional[Queue[int]]' = None) -> None:
    print(queue)
```

I'm using pyflakes 2.1.1 on Python 3.5.2 on Linux.

asottile:
Resolves #447

There's a couple cases that needed handling here:
- partially quoted annotations (as reported in the original issue)
- annotations which are doubly forward annotations (I don't have a word for this or a document that explains them, but `mypy` handles them fine -- things like `-> "foo['bar']"`
- quoted annotations while using `from __future__ import annotations`

### DIFF
--- a/pyflakes/checker.py
+++ b/pyflakes/checker.py
@@ -618,38 +618,53 @@ def getNodeName(node):
         return node.name
 
 
-def is_typing_overload(value, scope_stack):
-    def name_is_typing_overload(name):  # type: (str) -> bool
+def _is_typing(node, typing_attr, scope_stack):
+    def _bare_name_is_attr(name):
+        expected_typing_names = {
+            'typing.{}'.format(typing_attr),
+            'typing_extensions.{}'.format(typing_attr),
+        }
         for scope in reversed(scope_stack):
             if name in scope:
                 return (
                     isinstance(scope[name], ImportationFrom) and
-                    scope[name].fullName in (
-                        'typing.overload', 'typing_extensions.overload',
-                    )
+                    scope[name].fullName in expected_typing_names
                 )
 
         return False
 
-    def is_typing_overload_decorator(node):
-        return (
-            (
-                isinstance(node, ast.Name) and name_is_typing_overload(node.id)
-            ) or (
-                isinstance(node, ast.Attribute) and
-                isinstance(node.value, ast.Name) and
-                node.value.id == 'typing' and
-                node.attr == 'overload'
-            )
+    return (
+        (
+            isinstance(node, ast.Name) and
+            _bare_name_is_attr(node.id)
+        ) or (
+            isinstance(node, ast.Attribute) and
+            isinstance(node.value, ast.Name) and
+            node.value.id in {'typing', 'typing_extensions'} and
+            node.attr == typing_attr
         )
+    )
 
+
+def is_typing_overload(value, scope_stack):
     return (
         isinstance(value.source, FUNCTION_TYPES) and
         any(
-            is_typing_overload_decorator(dec)
+            _is_typing(dec, 'overload', scope_stack)
             for dec in value.source.decorator_list
         )
     )
+
+
+def in_annotation(func):
+    @functools.wraps(func)
+    def in_annotation_func(self, *args, **kwargs):
+        orig, self._in_annotation = self._in_annotation, True
+        try:
+            return func(self, *args, **kwargs)
+        finally:
+            self._in_annotation = orig
+    return in_annotation_func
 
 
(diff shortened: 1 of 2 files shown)


PeterJCLaw:
I've just run this against an internal codebase (~300k lines of Python) without errors. I don't think we had any of the cases which were previously erroring though, so take that with a pinch of salt.

asottile:
after chatting with some mypy devs, I need to additionally handle this case:

```python
Literal['foo', 'bar']
```

and PEP 593 is on the horizon, adding support for:

```python
Annotation[int, could(literally('be', 'anything'))]
```

so this is not quite ready yet

asottile:
I'm going to wait on implementing `Annotation` since it's a little more complicated and the PEP is only in the `Draft` state right now.  I've fixed (and tested) `Literal['foo', 'bar']`.  I've also put a place where it should be straightforward to extend for `Annotation` when `PEP 593` is ratified.

bluetech:
We've had several instances of this in pytest. With this branch, they are all fixed, except one inside an `attr.ib(type="...")`, but that's understandable. Didn't notice any other problems.

pyflakes/checker.py
```diff
@@ -1575,7 +1607,27 @@ def BINOP(self, node):
...
+                fn()
+            else:
+                self.deferFunction(fn)
+
+    if PY38_PLUS:
+        def CONSTANT(self, node):
+            if isinstance(node.value, str):
+                return self.STR(node)
+    else:
+        NUM = BYTES = ELLIPSIS = CONSTANT = ignore
```
PeterJCLaw: Just checking it's intentional that `NUM`, `BYTES` and `ELLIPSIS` aren't defined at all in the `PY38_PLUS` branch here? (I'm not sure I follow exactly what these methods do, just highlighting the imbalance) I'm guessing in later versions of Python they're all folded into `CONSTANT`?

asottile: yep, that was one of the major changes of python3.8's ast -- `ast.Constant` now represents all of the different types of constants

unmade:
After merging this into master, I get `F821 undefined name` for literal types, for example:

```python
def foo(bar: Literal["baz"]):
    ...

# F821 undefined name 'baz'
```

To reproduce install flake8 as follows:
```bash
pip install flake8 \
-e git+https://github.com/pycqa/pyflakes@c9708a18a17fbf17e0d88c1d1675ac1a926c4565#egg=pyflakes \
-e git+https://github.com/pycqa/pycodestyle@master#egg=pycodestyle
```

asottile:
@unmade can you show your full source?

asottile:
I ask because there's a test specifically for that: https://github.com/PyCQA/pyflakes/pull/479/files#diff-8dd6768783dc8a438172318bc691ac3eR457

unmade:
Sorry, I overlooked tests in PR, but I guess now I know what causes the trouble.

The problem is because I import `Literal` from another file instead of directly import from `typing` or `typing_extensions`, like this:

```python
from .compat import Literal
```

If I import `Literal` directly from typing, then everything is fine.

My full source:
- File I have trouble with:
    * https://github.com/unmade/apiwrappers/blob/master/src/apiwrappers/factories.py#L6
    * https://github.com/unmade/apiwrappers/blob/master/src/apiwrappers/factories.py#L19

- File from which I import `Literal`:
    * https://github.com/unmade/apiwrappers/blob/master/src/apiwrappers/compat.py

asottile:
yeah pyflakes doesn't do dynamic analysis so it can't follow imports like that

:thinking:

asottile:
@unmade could you take a look at #517 and see if it fixes your problem and if you think it is a good idea (if so, please review / approve the PR -- we have multiple sign off and it's often difficult to get reviewers!)

unmade:
@asottile Thanks for PR! It fixes my problem. I guess we can continue further discussions in #517

PR #484, opened 2019-10-24
Warn about is comparison to tuple

### LINKED ISSUE #483: python 3.8 SyntaxWarning: "is" with a literal. Did you mean "=="?
Hi there -

I searched a bit for this but couldn't find it.   This is a warning generated by Python 3.8 and I think it's new.  It's very hard to get pytest to raise for this, so it would be better if it were in pyflakes / flake8.

demo:

```
"""test a syntax warning."""

x = 5

if x is ():
    print("hi")

```

```
$ /opt/python3.8/bin/python3 -m py_compile test.py
test.py:5: SyntaxWarning: "is" with a literal. Did you mean "=="?
  if x is ():

```

asottile:
Resolves #483

### DIFF
--- a/pyflakes/checker.py
+++ b/pyflakes/checker.py
@@ -78,6 +78,51 @@ else:
     LOOP_TYPES = (ast.While, ast.For)
     FUNCTION_TYPES = (ast.FunctionDef,)
 
+
+if PY38_PLUS:
+    def _is_singleton(node):  # type: (ast.AST) -> bool
+        return (
+            isinstance(node, ast.Constant) and
+            isinstance(node.value, (bool, type(Ellipsis), type(None)))
+        )
+elif not PY2:
+    def _is_singleton(node):  # type: (ast.AST) -> bool
+        return isinstance(node, (ast.NameConstant, ast.Ellipsis))
+else:
+    def _is_singleton(node):  # type: (ast.AST) -> bool
+        return (
+            isinstance(node, ast.Name) and
+            node.id in {'True', 'False', 'Ellipsis', 'None'}
+        )
+
+
+def _is_tuple_constant(node):  # type: (ast.AST) -> bool
+    return (
+        isinstance(node, ast.Tuple) and
+        all(_is_constant(elt) for elt in node.elts)
+    )
+
+
+if PY38_PLUS:
+    def _is_constant(node):
+        return isinstance(node, ast.Constant) or _is_tuple_constant(node)
+else:
+    _const_tps = (ast.Str, ast.Num)
+    if not PY2:
+        _const_tps += (ast.Bytes,)
+
+    def _is_constant(node):
+        return (
+            isinstance(node, _const_tps) or
+            _is_singleton(node) or
+            _is_tuple_constant(node)
+        )
+
+
+def _is_const_non_singleton(node):  # type: (ast.AST) -> bool
+    return _is_constant(node) and not _is_singleton(node)
+
+
 # https://github.com/python/typed_ast/blob/1.4.0/ast27/Parser/tokenizer.c#L102-L104
 TYPE_COMMENT_RE = re.compile(r'^#\s*type:\s*')
 # https://github.com/python/typed_ast/blob/1.4.0/ast27/Parser/tokenizer.c#L1408-L1413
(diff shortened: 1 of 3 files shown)


pyflakes/test/test_is_literal.py
```diff
@@ -198,3 +198,25 @@ def test_chained_operators_is_str_end(self):
...
+            x = 5
+            if x is (1, '2', True, (1.5, ())):
+                pass
+        ''', IsLiteral)
+
+    def test_is_tuple_containing_variables_ok(self):
+        # a bit nonsensical, but does not trigger a SyntaxWarning
+        self.flakes('''\
+            x = 5
+            if x is (x,):
```
zzzeek: weird i wonder why this doesn't trigger and if that is intentional on python's end

bitglue:
Pyflakes doesn't otherwise check for syntax errors, since those get raised when generating the AST. Why is this one necessary to check specially?

pyflakes/checker.py
```diff
@@ -78,6 +78,51 @@ def getAlternatives(n):
...
+        isinstance(node, ast.Tuple) and
+        all(_is_constant(elt) for elt in node.elts)
+    )
+
+
+if PY38_PLUS:
+    def _is_constant(node):
+        return isinstance(node, ast.Constant) or _is_tuple_constant(node)
+else:
+    _const_tps = (ast.Str, ast.Num)
```
zzzeek: if py38 has solidified the idea of "constant" that might be why this syntaxwarning suddenly appeared

zzzeek:
> Pyflakes doesn't otherwise check for syntax errors, since those get raised when generating the AST. Why is this one necessary to check specially?

this error is not "raised" by the interpreter, it's given out as a warning, which is very difficult to detect in a test suite.   We raise exceptions on warnings however this particular one is still sneaking through before the filter gets set, apparently, as the condition here is virtually identical to existing checks that pyflakes performs it's appropriate that this one be present as well.

pyflakes/test/test_is_literal.py
asottile: this tuple isn't a constant literal 🤷‍♂

asottile:
> 
> 
> Pyflakes doesn't otherwise check for syntax errors, since those get raised when generating the AST. Why is this one necessary to check specially?

this isn't a `SyntaxError`, this is a warning in python3.8+ (afaict the check that's being augmented here was also in response to the new warning in python3.8)

pyflakes/checker.py
asottile: they're actually separate bpo issues:
- [folding into `ast.Constant` - bpo-32892](https://bugs.python.org/issue32892)
- [SyntaxWarning on `is` / `is not` compared with constant expressions - bpo-34850](https://bugs.python.org/issue34850)

pyflakes/messages.py
```diff
@@ -265,7 +265,7 @@ class InvalidPrintSyntax(Message):
 
 
 class IsLiteral(Message):
-    message = 'use ==/!= to compare str, bytes, and int literals'
+    message = 'use ==/!= to compare constant literals (str, bytes, int, float, tuple)'
```
asmeurer: For syntax errors pyflakes matches the interpreter error messages. Not sure if we should do the same here.

asottile: should I change it to:

`"is" / "is not" with a literal.  Did you mean "==" / "!="?`

```console
$ python3.8 -c '1 is 1'
<string>:1: SyntaxWarning: "is" with a literal. Did you mean "=="?
$ python3.8 -c '1 is not 1'
<string>:1: SyntaxWarning: "is not" with a literal. Did you mean "!="?
```

asmeurer: I don't know how important it is to make them match really.

pyflakes/test/test_is_literal.py
xmo-odoo: Maybe the trigger should be "anything other than a name or named constant"? It doesn't really make sense for non-constant literals to pass where constant literals fail, they're even more obviously incorrect.

asottile: the goal is to match the cpython warning

PR #490, opened 2019-11-19
Make pyflakes more resistant to syntax additions

asottile:
For example, if I comment out `JOINEDSTR`

```console
$ python3 -m pyflakes /dev/stdin <<< 'f"{1}"'
$ echo $?
0
$ PYFLAKES_ERROR_UNKNOWN=1 python3 -m pyflakes /dev/stdin <<< 'f"{1}"'
  File "/usr/lib/python3.6/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/usr/lib/python3.6/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/tmp/pyflakes/pyflakes/__main__.py", line 5, in <module>
    main(prog='pyflakes')
  File "/tmp/pyflakes/pyflakes/api.py", line 206, in main
    warnings = checkRecursive(args, reporter)
  File "/tmp/pyflakes/pyflakes/api.py", line 155, in checkRecursive
    warnings += checkPath(sourcePath, reporter)
  File "/tmp/pyflakes/pyflakes/api.py", line 99, in checkPath
    return check(codestr, filename, reporter)
  File "/tmp/pyflakes/pyflakes/api.py", line 74, in check
    w = checker.Checker(tree, file_tokens=file_tokens, filename=filename)
  File "/tmp/pyflakes/pyflakes/checker.py", line 771, in __init__
    self.handleChildren(tree)
  File "/tmp/pyflakes/pyflakes/checker.py", line 1198, in handleChildren
    self.handleNode(node, tree)
  File "/tmp/pyflakes/pyflakes/checker.py", line 1245, in handleNode
    handler(node)
  File "/tmp/pyflakes/pyflakes/checker.py", line 1198, in handleChildren
    self.handleNode(node, tree)
  File "/tmp/pyflakes/pyflakes/checker.py", line 1245, in handleNode
    handler(node)
  File "/tmp/pyflakes/pyflakes/checker.py", line 1023, in _unknown_handler
    raise NotImplementedError('Unexpected type: {}'.format(type(node)))
NotImplementedError: Unexpected type: <class '_ast.JoinedStr'>
```

this should _hopefully_ reduce the "could you make a release" comments that we get after every python version release

### DIFF
--- a/pyflakes/checker.py
+++ b/pyflakes/checker.py
@@ -1009,12 +1009,30 @@ class Checker(object):
 
         self.scope[value.name] = value
 
+    def _unknown_handler(self, node):
+        # this environment variable configures whether to error on unknown
+        # ast types.
+        #
+        # this is silent by default but the error is enabled for the pyflakes
+        # testsuite.
+        #
+        # this allows new syntax to be added to python without *requiring*
+        # changes from the pyflakes side.  but will still produce an error
+        # in the pyflakes testsuite (so more specific handling can be added if
+        # needed).
+        if os.environ.get('PYFLAKES_ERROR_UNKNOWN'):
+            raise NotImplementedError('Unexpected type: {}'.format(type(node)))
+        else:
+            self.handleChildren(node)
+
     def getNodeHandler(self, node_class):
         try:
             return self._nodeHandlers[node_class]
         except KeyError:
             nodeType = getNodeType(node_class)
-        self._nodeHandlers[node_class] = handler = getattr(self, nodeType)
+        self._nodeHandlers[node_class] = handler = getattr(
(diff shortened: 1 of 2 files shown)


myint:
Cool! Just curious if there is any risk of us getting bug reports about new syntax slower due to this.

asottile:
It depends, for the syntax changes where we need no special handling this'll just work for them.  when we need special handling presumably it would be ~slightly slower (since there wouldn't be a crash).  but I don't think it would slow our time-to-fix (since we'd presumably start testing it in pyflakes directly and notice it in our test suite)

my guess is that it wouldn't make a difference, but drastically improves the lifespan of a pyflakes release against syntax changes

PR #497, opened 2019-12-19
Allow complex expressions in the ignore part of a type comment

hczhcz:
Mypy allows several error codes with minus signs in them, like `# type: ignore[comparison-overlap]`. Also, it allows multiple error codes in one line, like `# type: ignore[name-defined, attr-defined]`. These cases will trigger syntax errors in pyflakes. I am not sure about whether there is a standard about the syntax. At least, we can make it compatible.

See also:
https://mypy.readthedocs.io/en/latest/error_code_list2.html
https://github.com/python/mypy/blob/master/test-data/unit/check-errorcodes.test

### DIFF
--- a/pyflakes/test/test_checker.py
+++ b/pyflakes/test/test_checker.py
@@ -155,6 +155,12 @@ class CollectTypeCommentsTests(TestCase):
     def test_type_comment_starts_with_word_ignore(self):
         ret = self._collect('x = 1 # type: ignore[T]')
         self.assertSetEqual(ret, set())
+        ret = self._collect('x = 1 # type: ignore[code-1]')
+        self.assertSetEqual(ret, set())
+        ret = self._collect('x = 1 # type: ignore[code-1, code-2]')
+        self.assertSetEqual(ret, set())
+        ret = self._collect('x = 1 # type: ignore[ code-1 , code-2 ]')
+        self.assertSetEqual(ret, set())
 
     def test_last_node_wins(self):
         """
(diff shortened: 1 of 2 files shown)


pyflakes/test/test_checker.py
```diff
@@ -155,6 +155,12 @@ def test_type_comment_without_whitespace(self):
     def test_type_comment_starts_with_word_ignore(self):
         ret = self._collect('x = 1 # type: ignore[T]')
         self.assertSetEqual(ret, set())
+        ret = self._collect('x = 1 # type: ignore[code-1]')
+        self.assertSetEqual(ret, set())
+        ret = self._collect('x = 1 # type: ignore[code-1, code-2]')
+        self.assertSetEqual(ret, set())
+        ret = self._collect('x = 1 # type: ignore[ code-1 , code-2 ]')
+        self.assertSetEqual(ret, set())
```
asottile: your tests pass without the change to the code

this is what `ASCII_NON_ALNUM` includes:

```
>>> pyflakes.checker.ASCII_NON_ALNUM
'\x00\x01\x02\x03\x04\x05\x06\x07\x08\t\n\x0b\x0c\r\x0e\x0f\x10\x11\x12\x13\x14\x15\x16\x17\x18\x19\x1a\x1b\x1c\x1d\x1e\x1f !"#$%&\'()*+,-./:;<=>?@[\\]^_`{|}~\x7f'
```

PR #499, opened 2020-01-01
Deprecate optparse in favor of argparse

sim-d:
According to [the official docs](https://docs.python.org/3.8/library/optparse.html), the optparse standard library has been deprecated since Python 3.2, in favor of the argparse standard library.

This pull request includes a commit that removes usage of optparse from pyflakes and uses argparse in its place. It also adds a couple of short descriptive strings for pyflakes usage via the command-line: a very brief one-liner describing pyflakes (taken from the README) and some information about the (optional) command-line arguments. Feel free to improve them, of course.

Note: this is my first fork, my first pull request, my first real usage of GitHub. I'm also relatively new to git and Python. Let me know if anything in my code or in the way I'm collaborating/etc. should be changed. I'm always trying to improve. (And happy new year!)

### DIFF
--- a/pyflakes/api.py
+++ b/pyflakes/api.py
@@ -201,7 +201,7 @@ def main(prog=None, args=None):
 
     parser = argparse.ArgumentParser(prog=prog,
                                      description='Check Python source files for errors')
-    parser.add_argument('--version', action='version', version=_get_version())
+    parser.add_argument('-V', '--version', action='version', version=_get_version())
     parser.add_argument('path', nargs='*',
                         help='Path(s) of Python file(s) to check. STDIN if not given.')
     args = parser.parse_args(args=args).path
--- a/pyflakes/api.py
+++ b/pyflakes/api.py
@@ -193,14 +193,18 @@ def _get_version():
 
 def main(prog=None, args=None):
     """Entry point for the script "pyflakes"."""
-    import optparse
+    import argparse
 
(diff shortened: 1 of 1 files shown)


sim-d:
And it looks like pull requests are based off of branches, not commits... Adding my recent commit (for a separate issue--#505) to the deprecate_optparse branch seemed to update #499. Sorry if I should have created a new branch for #505! If needed (e.g. if #505 should not be supported), how would I go about "removing" that recent commit? I know I could make a new commit which essentially un-does the update for #505, but I'm not sure if that's the best practice.
(Still learning.)

asottile:
ah, if you want to remove a commit you could do the following:

```bash
# make sure you're on the `deprecate_optparse` branch
git checkout deprecate_optparse
# *reset* back to the old commit
git reset --hard HEAD^  # here HEAD^ refers to the commit before the current revision, you could also explciitly use the exact hash
# then push this branch up -- note you need `--force` since you're erasing history
git push origin HEAD -f
```

in this case I think it's fine to keep the two commits together -- not sure how you'd do the second one without the first one (since it requires `argparse` features)

sim-d:
Thanks! Yeah, the recent commit is certainly dependent upon the prior, though I probably should have created a new branch for it so as not to interfere with this PR. I'll leave them together for now, but would be happy to remove the recent commit if the change proposed by #505 (and now included in this PR) is not desired.

PR #508, opened 2020-01-17
Fix annotation of posonlyarg

### LINKED ISSUE #507: F401 (imported but unused) false positive on Foo when using def x(a: Foo, /, *, b: Bar)
```python
from datetime import datetime as Foo, time as Bar


def x(a: Foo, /, *, b: Bar):
    pass
```

Running this on the file results in:

```
› flake8 --version  
3.7.9 (mccabe: 0.6.1, pycodestyle: 2.5.0, pyflakes: 2.1.1) CPython 3.8.1 on Linux
› pyflakes --version
2.1.1 Python 3.8.1 on Linux
› flake8 example.py 
example.py:1:1: F401 'datetime.datetime as Foo' imported but unused
› pyflakes example.py
example.py:1: 'datetime.datetime as Foo' imported but unused
```

asottile:
Resolves #507

### DIFF
--- a/pyflakes/checker.py
+++ b/pyflakes/checker.py
@@ -1799,6 +1799,10 @@ class Checker(object):
             addArgs(node.args.args)
             defaults = node.args.defaults
         else:
+            if PY38_PLUS:
+                for arg in node.args.posonlyargs:
+                    args.append(arg.arg)
+                    annotations.append(arg.annotation)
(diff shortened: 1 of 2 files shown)


mikeholler:
@asottile I've confirmed this one locally, so :+1:

adamstueckrath:
@asottile Also verified this works locally.

PR #511, opened 2020-02-14
Fix bug in async function scope checking

PeterJCLaw:
Previously attempting to check an async function scoped tree would immediately error because tuples are not callable.

The test added here is a direct copy of `test_scope_function` as I'm not sure what (if anything) should specifically be tested for async functions, though it serves to prove the fix. Happy to change this if there are other things we should be checking there.

### DIFF
--- a/pyflakes/test/test_code_segment.py
+++ b/pyflakes/test/test_code_segment.py
@@ -1,7 +1,9 @@
+from sys import version_info
+
 from pyflakes import messages as m
 from pyflakes.checker import (FunctionScope, ClassScope, ModuleScope,
                               Argument, FunctionDefinition, Assignment)
-from pyflakes.test.harness import TestCase
+from pyflakes.test.harness import TestCase, skipIf
 
 
 class TestCodeSegments(TestCase):
@@ -125,6 +127,7 @@ class TestCodeSegments(TestCase):
         self.assertIsInstance(function_scope_bar['h'], Argument)
         self.assertIsInstance(function_scope_bar['i'], Argument)
 
+    @skipIf(version_info < (3, 5), 'new in Python 3.5')
     def test_scope_async_function(self):
         checker = self.flakes('''
         async def foo(a, b=1, *d, **e):
--- a/pyflakes/test/test_code_segment.py
+++ b/pyflakes/test/test_code_segment.py
@@ -124,3 +124,49 @@ class TestCodeSegments(TestCase):
         self.assertIsInstance(function_scope_bar['g'], Argument)
         self.assertIsInstance(function_scope_bar['h'], Argument)
         self.assertIsInstance(function_scope_bar['i'], Argument)
+
+    def test_scope_async_function(self):
+        checker = self.flakes('''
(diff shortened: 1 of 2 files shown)


pyflakes/test/test_code_segment.py
```diff
@@ -124,3 +126,50 @@ def bar(f, g=1, *h, **i):
         self.assertIsInstance(function_scope_bar['g'], Argument)
         self.assertIsInstance(function_scope_bar['h'], Argument)
         self.assertIsInstance(function_scope_bar['i'], Argument)
+
+    @skipIf(version_info < (3, 5), 'new in Python 3.5')
+    def test_scope_async_function(self):
```
asottile: fwiw the other async functions live in `test/test_other.py` -- maybe move this function there?

on second look probably doesn't need the whole test that the other function scope test has -- just enough to demonstrate that functions nested in `async def` was broken

PeterJCLaw: Hrm, I considered `test/test_other.py`, but opted for this file because the issue here was with any code when the top level element in the tree was an async function. I don't think the signature or body of the function matters; perhaps I should remove that from the test to clarify it?

asottile: I guess I'm surprised this test didn't trigger the same bug -- so there must be something different -- https://github.com/PyCQA/pyflakes/blob/c9708a18a17fbf17e0d88c1d1675ac1a926c4565/pyflakes/test/test_other.py#L1924-L1929

also I think whatever test that demonstrates this error fits much better in the `TestAsyncStatements` test there

PeterJCLaw: Ah, I think the difference is the passing of `is_segment=True` -- that flag causes the harness to pick out the first child from the tree, meaning that the checker operates on a specific scope from the start rather than starting with a `ModuleScope` and then working its way down from there. (I initially couldn't work out why test in the commit which introduced this, and which is similar to `test_asyncDef`, didn't find the issue.)

The bug which this is fixing was to do with how `Checker.__init__` builds that root scope from the type of the initial tree node, so in most cases (which start with an `_ast.Module` and thus get a `ModuleScope`) we get away with it.

I'm guessing that this scenario would only be hit by a client using the `pyflakes` Python API directly -- certainly I couldn't construct any source which triggered it alone -- I spotted the bug from reading the code rather than experiencing the failure.

Happy to move the test and/or remove the new test and update that one if you'd prefer that, though my reasoning for the location of the test was that this issue relates to the segment behaviour as much as it does the fact that the function is async. What do you think?

asottile: ah ok, that makes way more sense now -- then this is the correct place :+1:

PeterJCLaw: :+1: Apologies for the lack of clarify upfront.

P.S: great gifs :)

asottile: hehe -- they come from a userscript https://github.com/chriskuehl/shipit

PR #512, opened 2020-02-14
Add a check for if statement conditions which are non-empty tuples

PeterJCLaw:
With the increasing prevalence of wrapping conditionals like:

``` python
    if (
        some_value != 'some_condition' and
        other_value is not 42
    )
```

and for wrapping other constructs with trailing commas, like:

``` python
    x = (
        some_value,
        other_value,
    )
```

it becomes quite easy to accidentally combine these into a conditional statement which is actually always true:

``` python
    if (
        some_value != 'some_condition' and
        other_value is not 42,
    )
```

This commit introduces a check for such constructs.

### DIFF
--- a/pyflakes/checker.py
+++ b/pyflakes/checker.py
@@ -1364,14 +1364,14 @@ class Checker(object):
         pass
 
     # "stmt" type nodes
-    DELETE = PRINT = FOR = ASYNCFOR = WHILE = IF = WITH = WITHITEM = \
+    DELETE = PRINT = FOR = ASYNCFOR = WHILE = WITH = WITHITEM = \
         ASYNCWITH = ASYNCWITHITEM = TRYFINALLY = EXEC = \
         EXPR = ASSIGN = handleChildren
 
     PASS = ignore
 
     # "expr" type nodes
-    BOOLOP = UNARYOP = IFEXP = SET = \
+    BOOLOP = UNARYOP = SET = \
         REPR = ATTRIBUTE = \
         STARRED = NAMECONSTANT = NAMEDEXPR = handleChildren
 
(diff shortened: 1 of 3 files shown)


pyflakes/messages.py
```diff
@@ -233,6 +233,13 @@ class TooManyExpressionsInStarredAssignment(Message):
     message = 'too many expressions in star-unpacking assignment'
 
 
+class IfTuple(Message):
+    """
+    Conditional test is a tuple, which are always True.
+    """
+    message = 'if condition is always true, perhaps remove parentheses?'
```
asottile: the messaging here probably needs to talk about it being a tuple -- removing parentheses doesn't fix it -- maybe something like '`if tuple literal` is always true, perhaps remove an accidental comma?'

PeterJCLaw: D'oh, this is me copying too much from the (otherwise fairly similar) `AssertTuple` case. That wording sounds good though :+1:

sigmavirus24:
I'm worried about false positives here but I'm also failing to think of any examples where they might be valid

asottile:
> 
> 
> I'm worried about false positives here but I'm also failing to think of any examples where they might be valid

I think any that would be problematic here would already be problematic for the existing `assert <tuple>` check -- so I imagine this to be pretty safe

pyflakes/messages.py
```diff
@@ -233,6 +233,13 @@ class TooManyExpressionsInStarredAssignment(Message):
     message = 'too many expressions in star-unpacking assignment'
 
 
+class IfTuple(Message):
+    """
+    Conditional test is a tuple, which are always True.
```
bluetech:
```suggestion
    Conditional test is a non-empty tuple literal, which are always True.
```

(The existing docstring for `AssertTuple` is inaccurate in the same way, too.)

PeterJCLaw: Nice :+1:

PeterJCLaw: I've separately put up https://github.com/PyCQA/pyflakes/pull/522 for `AssertTuple`

pyflakes/checker.py
```diff
@@ -1728,6 +1728,13 @@ def DICT(self, node):
                         )
         self.handleChildren(node)
 
+    def IF(self, node):
+        if isinstance(node.test, ast.Tuple) and node.test.elts != []:
```
asottile: I was going to suggest this change to `and node.test.elts` but I see below that `ASSERT` does the same thing

Issue #513, opened 2020-02-18
detect as unused imports subsumed by other imports


graingert:
*Please describe how you installed Flake8*

```
pip install flake8==3.7.9 pyflakes==2.1.1
```

```python
import sqlalchemy  # should be detected as unused
import sqlalchemy.exc


try:
    pass
except sqlalchemy.exc.DatabaseError:
    pass
```

asottile:
this is a duplicate of #429 / #137 (and a few others but those were the ones I could find quickly on my phone)

PR #514, opened 2020-02-18
add test for unused subsumed import

graingert:
see  #513


jayvdb (changes requested):
just noting that these tests intentionally fail - they are demos of what should be caught, but is not yet; CI doesnt show that.

jayvdb:
Maybe we should have a `test_failures.py` dedicated to known problems that need to get solved, but are known to be hard problems.

PR #516, opened 2020-02-23
Fix quoted type annotations in unusual contexts

### LINKED ISSUE #510: Spurious unsued import warning for type annotation used in quoted cast
Given
``` python
# demo.py 
from typing import Union, cast

x = cast('Union[str, int]', 42)
reveal_type(x)
```
Then I would hope that `pyflakes` would be able to see the usage of `Union` in the quoted cast, but it can't.
``` shell
$ pyflakes demo.py 
demo.py:1: 'typing.Union' imported but unused
demo.py:4: undefined name 'reveal_type'
```
(`reveal_type` being undefined is correct here).

`mypy` does seem happy with this usage though.
``` shell
$ mypy demo.py 
demo.py:4: error: Revealed type is 'Union[builtins.str, builtins.int]'
```

I realise this example is somewhat contrived (why would you quote this), however in the case that one of the union members is a circular import for example, you need to quote at least some part of the type and it's often cleaner to quote the whole name rather than parts of it.

PeterJCLaw:
This aims to fix #510.

There is a case which this doesn't fix, which might need a slightly different approach -- that of a quoted annotation within a generic subscript for a user-defined type (which is inherently not part of the `typing` module).
For this reason I'm tempted to break out the subscript case from the `cast` case. Currently this would be fairly easy to do (the `cast` case is handled by the first two commits here, the latter commits relate to the subscript case).

Do we want to solve the general generics case now, or is this useful enough as it is?
~If we do want to solve this now, suggestions on how to approach it would be useful. I can see a couple of routes:~
- ~try to walk the parents of the type we find to see if this subscript should be read as an annotation (likely somewhat involved because I don't think `pyflakes` has anything approaching this at the moment?)~
- ~try to parse the subscript _both_ as an annotation and otherwise; this feels like it could result in some false results, but might be easier to get working. If this approach is preferred, do we have some examples of this sort of thing?~

Conclusion: we'll ignore this case for now.

### DIFF
--- a/pyflakes/checker.py
+++ b/pyflakes/checker.py
@@ -674,7 +674,7 @@ def _is_typing_helper(node, is_name_match_fn, scope_stack):
     type annotation context.
 
     Note: you probably don't want to use this function directly. Instead see the
-    util below which wraps it (`_is_typing`).
+    utils below which wrap it (`_is_typing` and `_is_any_typing_member`).
     """
 
     def _bare_name_is_attr(name):
@@ -710,6 +710,16 @@ def _is_typing(node, typing_attr, scope_stack):
     context.
     """
     return _is_typing_helper(node, typing_attr.__eq__, scope_stack)
+
+
+def _is_any_typing_member(node, scope_stack):
+    """
+    Determine whether `node` represents any member of a typing module.
+
+    This is used as part of working out whether we are within a type annotation
+    context.
+    """
+    return _is_typing_helper(node, lambda x: True, scope_stack)
 
 
 def is_typing_overload(value, scope_stack):
@@ -1454,7 +1464,11 @@ class Checker(object):
             finally:
                 self._in_typing_literal = orig
         else:
-            self.handleChildren(node)
+            if _is_any_typing_member(node.value, self.scopeStack):
+                with self._enter_annotation():
+                    self.handleChildren(node)
+            else:
+                self.handleChildren(node)
 
     def _handle_string_dot_format(self, node):
         try:
--- a/pyflakes/checker.py
+++ b/pyflakes/checker.py
@@ -664,17 +664,26 @@ def getNodeName(node):
         return node.name
 
 
-def _is_typing(node, typing_attr, scope_stack):
+TYPING_MODULES = frozenset(('typing', 'typing_extensions'))
+
+
+def _is_typing_helper(node, is_name_match_fn, scope_stack):
+    """
+    Internal helper to determine whether or not something is a member of a
+    typing module. This is used as part of working out whether we are within a
+    type annotation context.
+
+    Note: you probably don't want to use this function directly. Instead see the
+    util below which wraps it (`_is_typing`).
+    """
+
     def _bare_name_is_attr(name):
-        expected_typing_names = {
-            'typing.{}'.format(typing_attr),
-            'typing_extensions.{}'.format(typing_attr),
-        }
         for scope in reversed(scope_stack):
             if name in scope:
                 return (
                     isinstance(scope[name], ImportationFrom) and
-                    scope[name].fullName in expected_typing_names
+                    scope[name].module in TYPING_MODULES and
+                    is_name_match_fn(scope[name].real_name)
                 )
 
         return False
@@ -686,10 +695,21 @@ def _is_typing(node, typing_attr, scope_stack):
         ) or (
             isinstance(node, ast.Attribute) and
             isinstance(node.value, ast.Name) and
-            node.value.id in {'typing', 'typing_extensions'} and
-            node.attr == typing_attr
+            node.value.id in TYPING_MODULES and
+            is_name_match_fn(node.attr)
         )
     )
+
+
+def _is_typing(node, typing_attr, scope_stack):
+    """
+    Determine whether `node` represents the member of a typing module specified
+    by `typing_attr`.
+
+    This is used as part of working out whether we are within a type annotation
+    context.
+    """
+    return _is_typing_helper(node, typing_attr.__eq__, scope_stack)
 
 
 def is_typing_overload(value, scope_stack):
--- a/pyflakes/checker.py
+++ b/pyflakes/checker.py
@@ -1563,6 +1563,15 @@ class Checker(object):
                 node.func.attr == 'format'
         ):
             self._handle_string_dot_format(node)
+
+        if (
+            _is_typing(node.func, 'cast', self.scopeStack) and
+            len(node.args) >= 1 and
+            isinstance(node.args[0], ast.Str)
+        ):
+            with self._enter_annotation():
+                self.handleNode(node.args[0], node)
+
         self.handleChildren(node)
 
     def _handle_percent_format(self, node):
--- a/pyflakes/checker.py
+++ b/pyflakes/checker.py
@@ -709,7 +709,7 @@ def _is_typing(node, typing_attr, scope_stack):
     This is used as part of working out whether we are within a type annotation
     context.
     """
-    return _is_typing_helper(node, typing_attr.__eq__, scope_stack)
+    return _is_typing_helper(node, lambda x: x == typing_attr, scope_stack)
 
 
 def _is_any_typing_member(node, scope_stack):
--- a/pyflakes/checker.py
+++ b/pyflakes/checker.py
@@ -8,6 +8,7 @@ import __future__
 import ast
 import bisect
 import collections
+import contextlib
 import doctest
 import functools
 import os
@@ -704,11 +705,8 @@ def is_typing_overload(value, scope_stack):
 def in_annotation(func):
     @functools.wraps(func)
     def in_annotation_func(self, *args, **kwargs):
-        orig, self._in_annotation = self._in_annotation, True
-        try:
+        with self._enter_annotation():
             return func(self, *args, **kwargs)
-        finally:
-            self._in_annotation = orig
     return in_annotation_func
 
 
@@ -1235,6 +1233,14 @@ class Checker(object):
                 del self.scope[name]
             except KeyError:
                 self.report(messages.UndefinedName, node, name)
+
+    @contextlib.contextmanager
+    def _enter_annotation(self):
+        orig, self._in_annotation = self._in_annotation, True
+        try:
+            yield
+        finally:
+            self._in_annotation = orig
 
     def _handle_type_comments(self, node):
         for (lineno, col_offset), comment in self._type_comments.get(node, ()):
--- a/pyflakes/test/test_type_annotations.py
+++ b/pyflakes/test/test_type_annotations.py
@@ -449,6 +449,23 @@ class TestTypeAnnotations(TestCase):
             return None
         """)
 
+    def test_quoted_type_cast(self):
+        self.flakes("""
+        from typing import cast, Optional
+
+        maybe_int = cast('Optional[int]', 42)
+        """)
+
+    def test_type_cast_literal_str_to_str(self):
+        # Checks that our handling of quoted type annotations in the first
+        # argument to `cast` doesn't cause issues when (only) the _second_
+        # argument is a literal str which looks a bit like a type annoation.
+        self.flakes("""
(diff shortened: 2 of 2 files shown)


pyflakes/checker.py
```diff
@@ -663,17 +664,27 @@ def getNodeName(node):
...
+class ANY_TYPING_NAME(object):
+    """
+    Marker for asking if the given node is a valid but unspeciffied typing member.
+
+    Note: this is *not* asking if the member is `typing.Any`.
+    """
+
+
 def _is_typing(node, typing_attr, scope_stack):
+    typing_modules = {'typing', 'typing_extensions'}
```
asottile: this would be slightly better as a module constant (so it doesn't rebuild the set on every call)

pyflakes/checker.py
```diff
@@ -663,17 +664,27 @@ def getNodeName(node):
...
 
+class ANY_TYPING_NAME(object):
+    """
+    Marker for asking if the given node is a valid but unspeciffied typing member.
+
+    Note: this is *not* asking if the member is `typing.Any`.
+    """
+
+
 def _is_typing(node, typing_attr, scope_stack):
```
asottile: the change to this function feels like it has been overloaded to do too much (over-DRYing) -- I would probably split the "ANY" concept to a separate function (it feels to me like the change here is complicating this function to do too much and it's harder to grok what's going on, whereas a separate function would be easier to comprehend)

put in another way, this function would be difficult to type-annotation (it would require a `Union`) which is often an indicator that it is doing too much

pyflakes/checker.py
```diff
@@ -1428,7 +1451,11 @@ def SUBSCRIPT(self, node):
             finally:
                 self._in_typing_literal = orig
         else:
-            self.handleChildren(node)
+            with contextlib.ExitStack() as stack:
```
asottile: ExitStack is python3-only :(

pyflakes/checker.py
```diff
@@ -1557,7 +1584,12 @@ def CALL(self, node):
                 node.func.attr == 'format'
         ):
             self._handle_string_dot_format(node)
-        self.handleChildren(node)
+
+        with contextlib.ExitStack() as stack:
+            if _is_typing(node.func, 'cast', self.scopeStack):
+                stack.enter_context(enter_annotation(self))
```
asottile: I believe this testcase will trigger a false positive:

```python
cast(str, 'Union[Unknown]')
```

(notably, only the first argument of `cast` is a typing context)

pyflakes/test/test_type_annotations.py
```diff
@@ -449,6 +449,31 @@ def f() -> Optional['Queue[str]']:
...
+        maybe_int = cast('Optional[int]', 42)
+        """)
+
+    @skipIf(version_info < (3,), 'new in Python 3')
+    def test_quoted_type_cast_renamed_import(self):
+        self.flakes("""
+        from typing import cast as tsac, Optional as Maybe
+
+        maybe_int = tsac('Maybe[int]', 42)
+        """)
```
asottile: all three of these tests should pass in python2.x (?)

pyflakes/checker.py
PeterJCLaw: D'oh. Good spot, will fix.

pyflakes/test/test_type_annotations.py
PeterJCLaw: Ah yeah. I'm too used to working Python 3 only.

PeterJCLaw:
@asottile thanks for the review. I can fix the issues you raised, though it would be good to get input on the approach for the subscript case as I think that's going to determine the best way to slice these changes between PRs -- do you have any suggestions on that?

pyflakes/test/test_type_annotations.py
asottile: same lol <3

asottile:
I *think* the only case that won't be caught for user-defined subscripted types is a subscript-at-typing-time-only type inside a user-defined subscripted type in a type alias -- and I can't really think of a heuristic to notice that :/

```python
class C(Generic[T]): ...

MyAlias = C['Queue[int]']
```

it's probably fine to ignore that edge case for now and we can circle back on it later, most of the common cases are caught because you can't have an alias that's purely a string (and commonly you'll be aliasing to some `typing`-module thing) and things inside variable annotations / function parameter annotations / return value annotations are well-known to be in a typing context

pyflakes/checker.py
PeterJCLaw: Done.

Fun fact: I've found that for small collections, comparing to membership of a `tuple` can sometimes actually be faster than sets, especially if the collection is declared inline. (Testing this case doesn't show a meaningful difference, so I've left it as a set.)

pyflakes/checker.py
PeterJCLaw: I've reworked this in 4037b061592b3f69ffbff8ee3c2a79f66d2ec175 so that we pass a name-comparison function in, rather than the target of the comparison. I'm not actually sure if that's better than just duplicating the function, so happy to change it if you think it's better to have the duplication.

pyflakes/test/test_type_annotations.py
PeterJCLaw: Done (with rebasing).

Meta: what's your (this repo/organisation)'s preference on rebasing versus adding commits to the end of the branch?

pyflakes/checker.py
PeterJCLaw: Fixed in 86e4e53bfbee362f044c52c72a552e07ec4ad5dd.

PeterJCLaw:
I realise I've not broken out the refactors yet; happy to do that once we're happy with the approach.

pyflakes/checker.py
```diff
@@ -663,17 +664,26 @@ def getNodeName(node):
         return node.name
 
 
-def _is_typing(node, typing_attr, scope_stack):
+TYPING_MODULES = {'typing', 'typing_extensions'}
```
asottile: probably `frozenset(('typing', 'typing_extensions'))` for immutability

pyflakes/checker.py
```diff
@@ -701,14 +732,20 @@ def is_typing_overload(value, scope_stack):
     )
 
 
+@contextlib.contextmanager
+def enter_annotation(instance):
+    orig, instance._in_annotation = instance._in_annotation, True
+    try:
+        yield
+    finally:
+        instance._in_annotation = orig
```
asottile: oops sorry should have mentioned this last review -- this probably makes more sense as a member function (that way `instance` can just be `self` -- less of a leaky abstraction that way too 👍)

pyflakes/test/test_type_annotations.py
asottile: that's hmmm a good question -- at the end we do merge commits but I think it's up to the author to tidy commits / etc.

pyflakes/checker.py
asottile: lg2m

pyflakes/checker.py
asottile: yep -- a tuple can be inlined into the code object (the peephole optimizer also does this for `in [1, 2, 3]` and `in {1, 2, 3}` as well -- converting the former to a tuple and the latter to a frozenset)

```pycon
>>> def f(x): return x in (1, 2, 3)
... 
>>> def g(x): return x in [1, 2, 3]
... 
>>> def h(x): return x in {1, 2, 3}
... 
>>> import dis
>>> dis.dis(f)
  1           0 LOAD_FAST                0 (x)
              2 LOAD_CONST               1 ((1, 2, 3))
              4 CONTAINS_OP              0
              6 RETURN_VALUE
>>> dis.dis(g)
  1           0 LOAD_FAST                0 (x)
              2 LOAD_CONST               1 ((1, 2, 3))
              4 CONTAINS_OP              0
              6 RETURN_VALUE
>>> dis.dis(h)
  1           0 LOAD_FAST                0 (x)
              2 LOAD_CONST               1 (frozenset({1, 2, 3}))
              4 CONTAINS_OP              0
              6 RETURN_VALUE
```

pyflakes/checker.py
```diff
@@ -1557,6 +1598,15 @@ def CALL(self, node):
...
         ):
             self._handle_string_dot_format(node)
+
+        if (
+            _is_typing(node.func, 'cast', self.scopeStack) and
+            len(node.args) >= 1 and
+            isinstance(node.args[0], ast.Str)
+        ):
+            with enter_annotation(self):
+                self.handleNode(node.args[0], node)
```
PeterJCLaw: I've just spotted that we do in fact a way to get the parent of a node, do you think it would be neater to push this into the `STR` method and have it look at the parent, rather than doing this (which potentially processes the `STR` node twice)?

(I'm assuming that the parent of the `ast.Str` node would end up being the call func, but haven't tested that)

asottile: I suspect that way would be more difficult to catch the `cast(str, 'not an annotation')` case -- this seems pretty striaghtforward as-is 🤷‍♂ -- up to you

PeterJCLaw: Ah, good point. Will leave as-is.

PeterJCLaw:
Rebased to squash together the refactors and separate them better from functional changes. Also took the opportunity to remove some of the dead changes (e.g: `ExitStack`).

PeterJCLaw:
Test case I've just thought of that I should add:
```
Func = Callable[
    ['Queue[int]', ...],
    None,
]
```
I _think_ the current implementation will catch this, but would be good to be sure.

jnsnow:
Hi, testing this on my local code fixes all the places where I had been seeing issues :)

tyvm!

pyflakes/checker.py
```diff
@@ -1557,6 +1597,15 @@ def CALL(self, node):
                 node.func.attr == 'format'
         ):
             self._handle_string_dot_format(node)
+
+        if (
+            _is_typing(node.func, 'cast', self.scopeStack) and
+            len(node.args) >= 1 and
```
bluetech: `typing.cast` takes exactly two arguments. Is the `>= 1` for robustness?

PeterJCLaw: Yeah -- it's to guard the `node.args[0]` below, rather than as part of the "is this `typing.cast` check.

pyflakes/checker.py
```diff
@@ -1557,6 +1597,15 @@ def CALL(self, node):
                 node.func.attr == 'format'
         ):
             self._handle_string_dot_format(node)
+
+        if (
```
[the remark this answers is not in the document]
PeterJCLaw: Huh, hadn't spotted that. I wonder if it's intentional that the other `if`s are double indented? @asottile maybe?

pyflakes/checker.py
```diff
@@ -1236,6 +1264,14 @@ def on_conditional_branch():
             except KeyError:
                 self.report(messages.UndefinedName, node, name)
 
+    @contextlib.contextmanager
+    def _enter_annotation(self):
```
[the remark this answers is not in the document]
PeterJCLaw: Nice idea. I'm going to leave this though as there's currently only one place which sets `self._in_typing_literal` and that code is untouched by this PR.

pyflakes/checker.py
```diff
@@ -685,12 +695,33 @@ def _bare_name_is_attr(name):
...
 
+def _is_typing(node, typing_attr, scope_stack):
+    """
+    Determine whether `node` represents the member of a typing module specified
+    by `typing_attr`.
+
+    This is used as part of working out whether we are within a type annotation
+    context.
+    """
+    return _is_typing_helper(node, typing_attr.__eq__, scope_stack)
```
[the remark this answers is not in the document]
PeterJCLaw: yeah, that is nicer: 0f73307

pyflakes/checker.py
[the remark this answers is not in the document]
asottile: if they blame to me it's probably habit -- it's so they're easier to differentiate from the code in the block

PR #517, opened 2020-03-02
Be more cautious when identifying typing Literal

asottile:
I feel that it's better to err on the side of false-negatives here than false-positives -- it's also especially unlikely that a typing-context subscript where the left hand side is "Literal" is anything other than `typing.Literal`

Originally brought up in https://github.com/PyCQA/pyflakes/pull/479#issuecomment-593150973

### DIFF
--- a/pyflakes/checker.py
+++ b/pyflakes/checker.py
@@ -1421,7 +1421,15 @@ class Checker(object):
         STARRED = NAMECONSTANT = NAMEDEXPR = handleChildren
 
     def SUBSCRIPT(self, node):
-        if _is_typing(node.value, 'Literal', self.scopeStack):
+        if (
+                (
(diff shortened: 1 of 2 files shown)


pyflakes/checker.py
```diff
@@ -1421,7 +1421,15 @@ def ignore(self, node):
         STARRED = NAMECONSTANT = NAMEDEXPR = handleChildren
 
     def SUBSCRIPT(self, node):
-        if _is_typing(node.value, 'Literal', self.scopeStack):
```
bluetech: Even though it's explained in the test, I think it'd be useful to have a comment here too explaining why it's not using `_is_typing`. Maybe in several years it will make sense to switch back.

PR #524, opened 2020-03-17
2.2.0

asottile:


asottile:
(I don't have pypi permission, I believe if this looks good @bitglue  or @florentx would need to promote this)

asottile:
#523 is needed for CI to be green

c1aris:
Hi @asottile ! Thanks for the great work. May I ask that will the latest commit "Fix pyflakes for removal of ast.Param" also be included in the 2.2.0a1 version? It seems that `ast.Param` is used in Python nightly in Linux as well.

asottile:
@c1aris it is, however it doesn't matter as they are reverting in cpython: https://github.com/python/cpython/pull/19056

c1aris:
Okay, thank you so much!

bitglue:
Let's just make it 2.2.0? There's always 2.2.1, if something needs fixing.

asottile:
sure, I figure an alpha release is probably better so there can be some soft vetting, but it's your project so I'm happy do what you want, I'll go drop the "a1"

bitglue:
Thanks for putting this together @asottile. And apologies to everyone else for the delay. I'm going to add some additional maintainers to pypi to busproof this process.

I had to tweak the version number in `__init__.py` so I amended the commit and merged c72d6cf1a9a119c1dd7a7674f36da21aea32d828.

bitglue:
Heh, I probably should have tweaked the release date in news.rst, too :(

The most tedious part of making these releases is updating that silly file. Would too many people be upset if we just deleted it? There's always the git history. Or perhaps there's some way to automatically generate it? Or we just require PRs to update it? Open to ideas.

asottile:
oops, I knew I forgot something 🤦 

the changelog is a bit tedious but I'm happy to fill it out when making releases 🤷  I think it's still somewhat valuable for large/popular projects such as pyflakes as traversing the git history on github is not always the easiest.  having contributors edit it often ends up in merge conflict messes unless you do something like towncrier to manage changelog fragments (and then merge them all together at the end) -- but then you end up being in the business of policing those which can be quite some work -- I guess worst case we continue what we're doing and I'm fine maintaining that

PR #530, opened 2020-04-15
Omit code guarded by 'if TYPE_CHECKING'

PetterS:
This commit finds a problem that has been observed in read code.

Consider the following code:

```
...
if TYPE_CHECKING
    from a import A, B
    from b import C

...
def f() -> "B":
    ...

def f()
    # Oops! C is acutally used here.
    C()
```

This commit ignores all code that is guarded by TYPE_CHECKING in
order to find mistakes like this. This constant is always False
when mypy is not running anyway.


asottile:
there's also `if False:` and `MYPY = False\nif MYPY:` and various other conventions for accomplishing the same thing

I don't think skipping the body there is quite right either as it would hide any errors in that block -- the block is not always just imports

your current patch breaks this for instance:

```python

if TYPE_CHECKING:
    T = ...

def f() -> 'T':
    pass
```

asottile:
I'm adding tests demonstrating that: ^ #531

PetterS:
There is also this case:
```python
from a import a

if TYPE_CHECKING:
    a()
```
Which probably should not raise an error?

asottile:
pyflakes currently doesn't do any branch analysis which in some cases resolves solving the halting problem

I'd argue that that code *should* raise an error, as it does during type checking time

PetterS:
> your current patch breaks this for instance:

This means the solution to this problem will have to be more complicated. Basically imports within the `TYPE_CHECKING` should be kept around in case they are used in string annotations, but ignored for all other purposes?

asottile:
Yes in some capacity, I also don't think this is limited to imports as any code can appear in such a block (functions, classes, etc.)

PetterS:
The updated code also passes with the tests added in #531.

pyflakes/checker.py
```diff
@@ -1160,11 +1162,15 @@ def handleNodeLoad(self, node):
                 # alias of other Importation and the alias
                 # is used, SubImportation also should be marked as used.
                 n = scope[name]
-                if isinstance(n, Importation) and n._has_alias():
-                    try:
-                        scope[n.fullName].used = (self.scope, node)
-                    except KeyError:
-                        pass
+                if isinstance(n, Importation):
```
asottile, line 1165: it's not just imports -- it's any code in there

for instance:

```python
if TYPE_CHECKING:
    class C:
        ...
```

pyflakes/checker.py
```diff
@@ -2120,6 +2131,7 @@ def IMPORT(self, node):
             else:
                 name = alias.asname or alias.name
                 importation = Importation(name, node, alias.name)
+            importation.during_type_checking = self._in_type_checking
```
asottile, line 2134: this should be part of the construction

asottile:
do you plan to handle `if False` / `if MYPY`? they're also common ways to approach this problem but I don't think `if False` is universally used and I don't think `if MYPY` is reliably detectable as being this particular case -- and the omission of handling those seems odd but I don't quite know the best thing to do there

pyflakes/checker.py
PetterS: Right, I wasn't too worried about that case, but let's af course fix that!

pyflakes/checker.py
PetterS: Will do! There is quite a hierarchy of classes that will touch but it makes sense I suppose.

PetterS:
> do you plan to handle if False / if MYPY? they're also common ways to approach this problem but I don't think if False is universally used and I don't think if MYPY is reliably detectable as being this particular case -- and the omission of handling those seems odd but I don't quite know the best thing to do there

I wasn't planning on handling those. ìf False` would be feasible I guess but as you say, `MYPY` could be hard to get right in all cases.

asottile:
> 
> 
> > do you plan to handle if False / if MYPY? they're also common ways to approach this problem but I don't think if False is universally used and I don't think if MYPY is reliably detectable as being this particular case -- and the omission of handling those seems odd but I don't quite know the best thing to do there
> 
> I wasn't planning on handling those. `if False` would be feasible I guess but as you say, `MYPY` could be hard to get right in all cases.

cool, just checking :)

PetterS:
Just double-checking: this is a feasible approach then?

Asking because adding `during_type_checking` to all `Binding` subclass constructors and handling all other types of bindings will expand the patch a bit.

pyflakes/checker.py
```diff
@@ -1072,6 +1074,8 @@ def addBinding(self, node, value):
                 break
         existing = scope.get(value.name)
 
+        value.during_type_checking = self._in_type_checking
```
asottile, line 1077: my comment before was because attribute stuffing is a smell (and a type checker such as mypy would error accordingly on this line), this should either be:

1. added to each of the types as part of their `__init__` signature
2. stored separately alongside the type (a tuple?)

and not tacked onto (~essentially immutable by interface) objects after construction

asottile: (another alternative)
3. stored in a separate structure

PetterS: Yeah, I agree. But I think we should store it separately somehow since otherwise we will have to change so many places where we create various bindings.

I think this is sound from a type-checking perspective, though? Since `value` is documented to be an instance of `Binding` we know that this property exists.

PetterS: @asottile Actually `Binding` has another attribute `used` that is used in the exact same way as `during_type_checking` currently is. So the more I think about this, the more I think the current way fits into the existing code.

PetterS: Uploaded an example. Not really sure about it, but at least it is another solution.

pyflakes/checker.py
```diff
@@ -588,6 +587,17 @@ def _add_to_names(container):
 
 class Scope(dict):
     importStarred = False       # set to True when import * is found
+    # Special key for checking whether a binding is defined only for type checking.
+    TYPE_CHECKING_ONLY = object()
+
+    def __init__(self):
+        super(Scope, self).__init__(self)
+        self[self.TYPE_CHECKING_ONLY] = collections.defaultdict(bool)
```
asottile, line 595: this is worse imo, this would be typed very strangely `Dict[Union[object, str], Union[Binding, Dict[str, bool]]` -- I think it would be better to have two stacks than this weirdness.  Also `Dict[str, bool]` is better represented as `Set[str]`

pyflakes/checker.py
```diff
@@ -588,6 +587,17 @@ def _add_to_names(container):
...
+
+    def items(self):
+        for key, val in super(Scope, self).items():
+            if key != self.TYPE_CHECKING_ONLY:
```
asottile, line 599: it hopefully doesn't matter since I don't think this solution is a good one, but when comparing against singletons you'd use `is not` / `is`

PetterS:
- Having a separate stack would mean a lot of work keeping track of both stacks when creating scopes and iterating over the stacks. I don't see how this could be done in a good way. I am not at this point comfortable with any large refactors to this class.
- Adding a parameter to the constructors of all `Binding` subclasses would mean a lot of places suddenly needing to know about `during_type_checking`. Not a good solution either.

I think it is really good to keep this new functionality in a single place. But at this point I almost feel that the current `during_type_checking` attribute on `Binding` is the cleanest solution. It follows `used` in the sense that it is a mutable property, albeit not one meant to be changed after the binding is put into the `scope`.

Here is another solution that could be acceptable:
- Add the parameter to all `Binding` constructors. All different `Binding`s are then created in a new method:
  ```
  def createBinding(self, klass, *args, **kwargs):
    return klass(*args, during_type_checking=self._in_type_checking, **kwargs)
  ```
  so instead of e.g.
  ```
  importation = SubmoduleImportation(alias.name, node)
  ```
  we would write
  ```
  importation = self.createBinding(SubmoduleImportation, alias.name, node)
  ```

PetterS:
@asottile 

> added to each of the types as part of their __init__ signature

The current version of this PR now does this.

pyflakes/test/test_imports.py
```diff
@@ -1031,6 +1039,36 @@ def test_futureImportStar(self):
...
+        self.flakes('''
+        from typing import TYPE_CHECKING
+        if TYPE_CHECKING:
+            class T:
+                pass
+        t = T()
+        ''', m.UndefinedName)
+
+    @skipIf(version_info < (3,), 'has type annotations')
+    def test_usesTypingImportsForAnnotations(self):
```
asottile, line 1062: fwiw, the new tests are using `snake_case`

pyflakes/test/test_imports.py
```diff
@@ -1031,6 +1039,36 @@ def test_futureImportStar(self):
...
+        """Ignores definitions within 'if TYPE_CHECKING' checking normal code."""
+        self.flakes('''
+        from typing import TYPE_CHECKING
+        if TYPE_CHECKING:
+            class T:
+                pass
+        t = T()
+        ''', m.UndefinedName)
+
+    @skipIf(version_info < (3,), 'has type annotations')
```
asottile, line 1061: these can use `# type: () -> b` comments such that this is validated in python2 and python3

pyflakes/checker.py
```diff
@@ -310,10 +310,12 @@ class Binding(object):
                 the node that this binding was last used.
     """
 
-    def __init__(self, name, source):
+    def __init__(self, name, source, during_type_checking):
         self.name = name
         self.source = source
         self.used = False
+        assert isinstance(during_type_checking, bool)
+        self.during_type_checking = during_type_checking
```
asottile, line 318: maybe `for_annotations` is a more descriptive name here?  (idk, I'm bikeshedding on this, feel free to ignore this suggestion, it just feels a little like the meaning of the name gets lost a bit)

PetterS: It's fine; takes a couple of seconds to change.

pyflakes/test/test_imports.py
PetterS: Done.

pyflakes/test/test_imports.py
PetterS: I can't get this to work. I add an `assert False` right after the `continue` in `_collect_type_comments` and it never fires in the test (after I change to a ` # type: () -> b` comment)

PetterS: Oh! It's the `test_doctests` that runs the tests again with some extra stuff that breaks type comments. :-(

At least I can fix it now

asottile:
this regresses the following code sample:

```python


if TYPE_CHECKING:
    from typing import TypedDict
else:
    TypedDict = dict


class C(TypedDict):
    x: int
```

PetterS:
Will continue working on this PR when we have CI up and running. We need the following PRs:
- https://github.com/PyCQA/pyflakes/pull/532
- https://github.com/PyCQA/pyflakes/pull/531

asottile:
@PetterS sounds good -- we have a dual sign off and if you feel like you have enough context feel free to approve PRs.  I've gone ahead and merged #532 since it's going to block other things anyway.  You might also be interested in the the different approach taken by @srittau in #535 -- I hadn't considered an approach like that and it might just work perfectly for your thing as well!

PetterS:
Updated this PR with the new tests.

pyflakes/checker.py
```diff
@@ -1833,7 +1848,20 @@ def DICT(self, node):
     def IF(self, node):
         if isinstance(node.test, ast.Tuple) and node.test.elts != []:
             self.report(messages.IfTuple, node)
-        self.handleChildren(node)
+
+        prev = self._in_type_checking
+        if _is_typing(node.test, 'TYPE_CHECKING', self.scopeStack):
+            self._in_type_checking = True
+        # Handle the body of this if statement.
+        body_nodes = node.body
```
PetterS, line 1856: We could also add another argument only=["body"] to handleChildren to make it clearer.

PetterS:
@asottile merged to the latest master.

PetterS:
@asottile Should I merge this to master again? Is this likely to be merged?

pyflakes/checker.py
```diff
@@ -1973,7 +1988,20 @@ def DICT(self, node):
...
+        if not isinstance(body_nodes, list):
+            body_nodes = [body_nodes]
```
asottile, line 1998: does this code ever run? I think the body is always a list here (?)

pyflakes/test/test_imports.py
```diff
@@ -1031,6 +1039,48 @@ def test_futureImportStar(self):
         from __future__ import *
         ''', m.FutureFeatureNotDefined)
 
+    def test_ignores_typing_imports(self):
```
asottile, line 1042: let's put these tests into the type_annotations test

pyflakes/test/test_type_annotations.py
```diff
@@ -704,6 +704,27 @@ def f():  # type: () -> int
                     pass
         """)
 
+    def test_typing_guard_with_elif_branch(self):
+        # This test will actually not raise an error, even though it by analysis can
+        # be shown to have an error (Protocol is not defined outside TYPE_CHECKING).
+        # Pyflakes currently does not to case analysis.
```
asottile, line 710: to => do

asottile (approved):
sorry I lapsed on reviewing this -- looks good, just some very minor comments

pyflakes/checker.py
```diff
@@ -1216,6 +1226,10 @@ def handleNodeLoad(self, node):
                         scope[n.fullName].used = (self.scope, node)
                     except KeyError:
                         pass
+                if n.for_annotations and not self._in_annotation:
```
terencehonles, line 1229: Should this not happen before `scope[name].used` is marked? It's not really being used and the variable should be eligible to be released.

See: https://github.com/PyCQA/pyflakes/pull/622/files#diff-5df44d79406311387e0056faae7a8092d1ce9cc85c7d15420186e45a0c712283L1204-R1218

pyflakes/checker.py
terencehonles: in `IFEXPR` body is not a list.

terencehonles (commented):
I believe the main thing compared to #622 is this doesn't support `if not TYPE_CHECKING`, but that's probably not as likely as the other way so it's probably fine to not handle that.

PetterS:
Thanks for the comments! Now I have to remember what I wrote before! 😄  Will try to update the PR soon.

pyflakes/checker.py
PetterS: Yes, there is a test that will fail if this if statement is removed.

pyflakes/test/test_imports.py
PetterS: Done!

pyflakes/test/test_type_annotations.py
PetterS: Done!

PetterS:
@asottile made changes based on your review now.

pyflakes/checker.py
PetterS: If we do this, the following test will change:
```python
if TYPE_CHECKING:                                                                                                                                                                              
    from a import b
b()
```
It will also return `'a.b' imported but unused` after this change. Perhaps that is better.

We can do this.

PetterS: Done. Thanks.

PetterS:
@asottile ping on this.

asottile:
> 
> 
> @asottile ping on this.

I already approved, pyflakes (unfortunately) has multi-sign-off

asottile:
> 
> 
> @asottile is there anyone specific that should be signing off?

I've been going with the loosest interpretation of that possible and haven't had any pushback -- that is "anyone" + "a maintainer" being 2

srittau (approved):
LGTM, although I would personally default the `for_annotations` to `False`. Especially the test code would be a bit cleaner.

pyflakes/checker.py
```diff
@@ -321,10 +321,12 @@ class Binding(object):
                 the node that this binding was last used.
     """
 
-    def __init__(self, name, source):
+    def __init__(self, name, source, for_annotations):
         self.name = name
         self.source = source
         self.used = False
+        assert isinstance(for_annotations, bool)
```
sigmavirus24, line 328: What's the purpose of this assert here?

terencehonles:
> LGTM, although I would personally default the `for_annotations` to `False`. Especially the test code would be a bit cleaner.

I +1 this suggestion and what I had done in my copy of this same effort #622. I didn't bring it up because it's kinda a nit, but since the comment has been made I'll let @PetterS know there is at least one other person who was biting their tongue before :sweat_smile:

PetterS:
> > @asottile is there anyone specific that should be signing off?
> 
> I've been going with the loosest interpretation of that possible and haven't had any pushback -- that is "anyone" + "a maintainer" being 2

Great, let's merge it then!

asottile:
> 
> 
> > > @asottile is there anyone specific that should be signing off?
> > 
> > 
> > I've been going with the loosest interpretation of that possible and haven't had any pushback -- that is "anyone" + "a maintainer" being 2
> 
> Great, let's merge it then!

wellll now there's two open discussions that should probably be decided / resolved first

rggjan:
Any updates on merging this PR?

asottile:
> Any updates on merging this PR?

just checking:

are you able to see the merge conflict message?

screenshot of merge conflict

and the comment about unresolved discussions?

screenshot of my comment

rggjan:
> are you able to see the merge conflict message?

Actually, I missed this when focusing on the big green checkmarks

Screenshot 2022-09-21 at 15 47 11

> and the comment about unresolved discussions?

Yes, and I assumed it was related to the Merged and Closed point here:

Screenshot 2022-09-21 at 15 46 27

Thanks for the hints, though :)

asottile:
> Yes, and I assumed it was related to the Merged and Closed point here:
> 
> Screenshot 2022-09-21 at 15 46 27
> 
> Thanks for the hints, though :)

easy to click through and see one of those is closing an issue as a duplicate and the other is a PR on an unrelated project!

joaoe:
Howdy. Just a friendly reminder that is is sorely needed :) Can anyone rebase and merge ? Thanks.

PetterS:
I will not be updating this PR.

terencehonles:
I haven't been following this PR in awhile, but I've been making updates to my PR https://github.com/PyCQA/pyflakes/pull/622 since we're using it at my work and I just updated it again now and saw this conversation.

I believe it addresses all comments that were raised in this PR which was developed independently

PR #532, opened 2020-04-19
Fix test for nightly Python (3.9)

PetterS:
Trying to get the nightly test in #530 to pass.

### DIFF
--- a/pyflakes/test/test_api.py
+++ b/pyflakes/test/test_api.py
@@ -449,6 +449,8 @@ def baz():
         with self.makeTempFile(source) as sourcePath:
             if PYPY:
                 message = 'end of file (EOF) while scanning triple-quoted string literal'
+            elif sys.version_info >= (3, 9):
+                message = 'invalid string prefix'
             else:
                 message = 'invalid syntax'
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,5 +12,10 @@ matrix:
       dist: xenial
     - python: nightly
       dist: xenial
+# TODO: https://bugs.python.org/issue40334
+# the PEG parser is currently broken in some cases
+  allow_failures:
+    - python: nightly
+      dist: xenial
 install: pip install tox
 script: tox -e py


PetterS:
OK, test passes now but feels way too brittle.

Should be enough to assert that the error starts with `sourcePath + ":8:"`

pyflakes/test/test_api.py
```diff
@@ -447,19 +447,26 @@ def evaluate(source):
...
+            column = 11
+            whitespace_before_caret = 10
             if PYPY:
                 message = 'end of file (EOF) while scanning triple-quoted string literal'
-            else:
-                message = 'invalid syntax'
+            elif sys.version_info >= (3, 9):
+                message = 'invalid string prefix'
+                column = 52
+                whitespace_before_caret = 14
```
asottile, line 458: these offsets and error message seem like a bug, could you report this to upstream?  there is no line 14 and column 52 for starters and it's unclear why it's talking about a string prefix when the error is an unterminated string literal

PetterS: The reason it is talking about a string prefix is that that's the first error in encounters. The invalid string prefix is "quux".

PetterS: I don't know what to report to upstream. When I run the code, Python 3.9.0a5 gives me
```
  File "/mnt/c/Users/Admin/Desktop/syntaxerror.py", line 6
    '''quux'''
          ^
SyntaxError: invalid string prefix
```
which is correct.

And if I catch as print the exc_info I get
```
(<class 'SyntaxError'>, SyntaxError('invalid string prefix', ('<string>', 7, 50, "    '''\ndef bar():\n    pass\ndef baz():\n    '''quux'''\n")), <traceback object at 0x7fb1c824b340>)
```
This is identical to the system Python 3.6

I have no idea what the tests in this file are for, so can't have an opinion about what is reasonable here.

asottile: that's fine, I've gone ahead and made the report myself :) https://bugs.python.org/issue40335

PetterS: Aha! That's why I got the same with Python 3.6. Should have tried 3.8 as well

asottile:
could you rebase this? it should be fixed upstream now

[here's a simpler patch you can use instead](https://github.com/python/cpython/pull/19619#issuecomment-616659821)

PetterS:
Done!

But the Travis check does not seem to have run for some reason.

asottile:
looks like the new PEG parser has broken these even further :(

PetterS:
Drop these tests for nightly? Seems a little too fragile?

asottile:
the tests are valuable, and have pointed out a bug in cpython

I think we'll probably `allow_failures: ...` for now though to allow patches to flow while we wait for cpython to fix things

PetterS:
OK, where is that setting?

asottile:
https://docs.travis-ci.com/user/customizing-the-build/#jobs-that-are-allowed-to-fail

PR #534, opened 2020-04-24
Fix quoted type annotations in typing.TypeVar contexts

bluetech:
[`TypeVar`](https://docs.python.org/3/library/typing.html#typing.TypeVar) has such contexts in the positional arguments (for constraints) and in the `bound` keyword argument.

### DIFF
--- a/pyflakes/checker.py
+++ b/pyflakes/checker.py
@@ -1614,6 +1614,19 @@ class Checker(object):
             with self._enter_annotation():
                 self.handleNode(node.args[0], node)
 
+        elif _is_typing(node.func, 'TypeVar', self.scopeStack):
+            # TypeVar("T", "int", "str")
+            for arg in node.args[1:]:
+                if isinstance(arg, ast.Str):
+                    with self._enter_annotation():
+                        self.handleNode(arg, node)
+
+            # TypeVar("T", bound="str")
+            for keyword in node.keywords:
+                if keyword.arg == 'bound' and isinstance(keyword.value, ast.Str):
+                    with self._enter_annotation():
+                        self.handleNode(keyword.value, node)
+
         self.handleChildren(node)
 
     def _handle_percent_format(self, node):
--- a/pyflakes/test/test_type_annotations.py
+++ b/pyflakes/test/test_type_annotations.py
@@ -489,6 +489,21 @@ class TestTypeAnnotations(TestCase):
         maybe_int = tsac('Maybe[int]', 42)
         """)
 
+    def test_quoted_TypeVar_constraints(self):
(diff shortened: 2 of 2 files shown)


pyflakes/checker.py
```diff
@@ -1614,6 +1614,19 @@ def CALL(self, node):
             with self._enter_annotation():
                 self.handleNode(node.args[0], node)
 
+        if _is_typing(node.func, 'TypeVar', self.scopeStack):
```
asottile, line 1617:
```suggestion
        elif _is_typing(node.func, 'TypeVar', self.scopeStack):
```

pyflakes/checker.py
```diff
@@ -1614,6 +1614,19 @@ def CALL(self, node):
...
+            # TypeVar("T", "int", "str")
+            for arg in node.args[1:]:
```
asottile, line 1619: ah clever, no need to check the arg length when there's a slice

```pycon
>>> [][1:]
[]
```

pyflakes/checker.py
bluetech: Changed to `elif`, thanks.

asottile:
Validated this against a bunch of code, overall seemed to not regress anything, just this one improvement from the codebases I surveyed:

```diff
--- out_master/pandas.log	2021-01-04 13:33:02.789490134 -0800
+++ out_pr2/pandas.log	2021-01-04 13:35:34.541493678 -0800
@@ -179,8 +179,6 @@
 ./pandas/_testing/__init__.py:81:1 'pandas._testing.contexts.set_timezone' imported but unused
 ./pandas/_testing/__init__.py:81:1 'pandas._testing.contexts.use_numexpr' imported but unused
 ./pandas/_testing/__init__.py:81:1 'pandas._testing.contexts.with_csv_dialect' imported but unused
-./pandas/_typing.py:37:5 'pandas.core.arrays.base.ExtensionArray' imported but unused
-./pandas/_typing.py:39:5 'pandas.core.generic.NDFrame' imported but unused
 ./pandas/api/__init__.py:2:1 'pandas.api.extensions' imported but unused
 ./pandas/api/__init__.py:2:1 'pandas.api.indexers' imported but unused
 ./pandas/api/__init__.py:2:1 'pandas.api.types' imported but unused
```

Those types were used only in `TypeVar` contexts and had `noqa` on them already

PeterJCLaw:
Thanks for fixing this :) Do you know when it's likely to make it into a release?

asottile:
@PeterJCLaw please see the pinned issue

PeterJCLaw:
d'oh, sorry.

PR #535, opened 2020-04-26
Fix undefined name in annotations

### LINKED ISSUE #486: Undefined name when using variable annotations
Please consider the following file:

```python
Type: object
def foo(x: Type) -> None: ...
```

When checking it with '2.1.1 Python 3.8.0 on Linux' (or Python 3.7.3) I get the following error:

```
foo.py:2: undefined name 'Type'
```

This is problematic when checking python stub files.

srittau:
Variable annotations without a value don't create a name, but they still
can be used as variable annotation themselves as long as the annotation
is quoted or "from __future__ import annotations" is used.

The implementation introduces a new binding "Annotation" for these kinds
of variable annotations. This can potentially be used for further
annotation-related checks in the future.

Annotation handling is extended to be able to detect both forms of postponed
annotations (quoted and future import) during checks.

Fixes #486

### DIFF
--- a/pyflakes/checker.py
+++ b/pyflakes/checker.py
@@ -80,7 +80,7 @@ else:
     FUNCTION_TYPES = (ast.FunctionDef,)
 
 if PY36_PLUS:
-    ANNASSIGN_TYPES = (ast.AnnAssign)
+    ANNASSIGN_TYPES = (ast.AnnAssign,)
 else:
     ANNASSIGN_TYPES = ()
 
--- a/pyflakes/checker.py
+++ b/pyflakes/checker.py
@@ -746,6 +746,12 @@ def is_typing_overload(value, scope_stack):
     )
 
 
+class AnnotationState:
+    NONE = 0
+    STRING = 1
+    BARE = 2
+
+
 def in_annotation(func):
     @functools.wraps(func)
     def in_annotation_func(self, *args, **kwargs):
@@ -757,7 +763,7 @@ def in_annotation(func):
 def in_string_annotation(func):
     @functools.wraps(func)
     def in_annotation_func(self, *args, **kwargs):
-        with self._enter_annotation(string=True):
+        with self._enter_annotation(AnnotationState.STRING):
             return func(self, *args, **kwargs)
     return in_annotation_func
 
@@ -848,7 +854,7 @@ class Checker(object):
     nodeDepth = 0
     offset = None
     traceTree = False
-    _in_annotation = None  # other possible values: "bare", "string"
+    _in_annotation = AnnotationState.NONE
     _in_typing_literal = False
     _in_deferred = False
 
@@ -1292,8 +1298,8 @@ class Checker(object):
                 self.report(messages.UndefinedName, node, name)
 
     @contextlib.contextmanager
-    def _enter_annotation(self, string=False):
-        orig, self._in_annotation = self._in_annotation, "string" if string else "bare"
+    def _enter_annotation(self, ann_type=AnnotationState.BARE):
+        orig, self._in_annotation = self._in_annotation, ann_type
         try:
             yield
         finally:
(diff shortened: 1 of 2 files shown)


srittau:
The build failures (on Python nightly) seem unrelated to this PR.

pyflakes/checker.py
```diff
@@ -79,6 +79,10 @@ def getAlternatives(n):
     LOOP_TYPES = (ast.While, ast.For)
     FUNCTION_TYPES = (ast.FunctionDef,)
 
+if PY36_PLUS:
+    ANNASSIGN_TYPES = (ast.AnnAssign)
```
asottile, line 83: I assume you mean `(ast.AnnAssign,)`

pyflakes/checker.py
```diff
@@ -826,7 +848,7 @@ class Checker(object):
     nodeDepth = 0
     offset = None
     traceTree = False
-    _in_annotation = False
+    _in_annotation = None  # other possible values: "bare", "string"
```
asottile, line 851: I would prefer something a little less stringly typed 🤔 -- an enum would be appropriate but we need to support python 2 without dependencies (feel free to ignore -- I can't think of a better way to do this given the constraints)

pyflakes/checker.py
```diff
@@ -1265,13 +1292,17 @@ def on_conditional_branch():
                 self.report(messages.UndefinedName, node, name)
 
     @contextlib.contextmanager
-    def _enter_annotation(self):
-        orig, self._in_annotation = self._in_annotation, True
+    def _enter_annotation(self, string=False):
+        orig, self._in_annotation = self._in_annotation, "string" if string else "bare"
```
asottile, line 1296: could be `ann_type='bare'` and then no need for the boolean

pyflakes/test/test_type_annotations.py
```diff
@@ -300,6 +308,13 @@ class A:
         class B: pass
         ''', m.UndefinedName)
 
+        self.flakes('''
+        from __future__ import annotations
+        T: object
+        def f(t: T): pass
+        def g(t: 'T'): pass
+        ''')
+
```
asottile, line 317: one test I'd be interested in is:

```python
def f():
    x_is_unused: int
```

pyflakes/checker.py
srittau: Good catch, will fix.

pyflakes/checker.py
srittau: Other ideas I have: Use `""` instead of `None`. Or use numeric constants as enum simulation.

asottile: oh numeric might be nice, something like:
```python
class AnnotationState:
    NONE = 0
    STRING = 1
    BARE = 2
```

and then when dropping python2 it could be converted to a first-class enum

pyflakes/test/test_type_annotations.py
srittau: I have added this test now. This is actually something I noticed while implementing this. Would it make sense to change the current behavior (no error) in a future PR to flag this? Outside modules and class definitions this looks like an oversight.

asottile: hmmm interesting, yeah I was hoping it would flag it similar to the unused local

```console
$ ~/opt/venv/bin/pyflakes
def f():
    x_is_unused = 1
<stdin>:2: local variable 'x_is_unused' is assigned to but never used
```

but afaict it's not a regression and can be done later

___

I've definitely seen cases where annotate-then-assign is helpful, especially with really long types so it would only really make sense if it's annotated and then never used

```python
self.thing: Dict[str, Set[Tuple[int, str]]]
self.thing = collections.defaultdict(set)
```

asottile: so maybe a TODO comment that this should be improved in the future 👍

srittau: I have added a few more tests and a TODO comment.

asottile:
a quick rebase should make travis green -- I've marked nightly as `allow_failures: true` for now

sigmavirus24 (approved):

PR #580, opened 2020-09-27
fix: don't return syntax error for Annotated's metadata

### LINKED ISSUE #574: syntax error in forward annotation when using Annotated
When using [`Annotated`](https://www.python.org/dev/peps/pep-0593/) with string annotations the following error is raised:

```
./pyflakes.py:3:16 syntax error in forward annotation '>1'
```

Triggering code:

```
from typing import Annotated

Annotated[int, '>1']
```

sk-:
PEP 593 https://www.python.org/dev/peps/pep-0593/ introduced the type Annotated, which allows to add metadata to types. By the definition only the first argument is required to be a proper type, all the other arguments are dependent on the consumer and can be other types or even literals.

This PR fixes #574.

### DIFF
--- a/pyflakes/checker.py
+++ b/pyflakes/checker.py
@@ -827,6 +827,7 @@ class Checker(object):
     offset = None
     traceTree = False
     _in_annotation = False
+    _in_typing_annotated = False
     _in_typing_literal = False
     _in_deferred = False
 
@@ -1271,6 +1272,14 @@ class Checker(object):
             yield
         finally:
             self._in_annotation = orig
+    
+    @contextlib.contextmanager
+    def _enter_typing_annotated(self):
+        orig, self._in_typing_annotated = self._in_typing_annotated, True
+        try:
+            yield
+        finally:
+            self._in_typing_annotated = orig
 
     def _handle_type_comments(self, node):
         for (lineno, col_offset), comment in self._type_comments.get(node, ()):
@@ -1471,6 +1480,33 @@ class Checker(object):
                 self.handleChildren(node)
             finally:
                 self._in_typing_literal = orig
+        elif (
+                (
+                    isinstance(node.value, ast.Name) and
+                    node.value.id == 'Annotated'
+                ) or (
+                    isinstance(node.value, ast.Attribute) and
+                    node.value.attr == 'Annotated'
+                )
+        ):
+            orig = self._in_typing_annotated
+            self.handleNode(node.value, node)
+
+            if (
+                isinstance(node.slice, ast.Index) and
+                isinstance(node.slice.value, ast.Tuple)):
+                slice_children = list(iter_child_nodes(node.slice.value))
+                if slice_children:
+                    self.handleNode(slice_children[0], node.slice)
+                # Only the first argument of an Annotated type will always
+                # correspond to an actual type.
+                with self._enter_typing_annotated():
+                    for slice_child in slice_children[1:]:
+                        self.handleNode(slice_child, node)
+            else:
+                self.handleNode(node.slice, node)
+
+            self.handleNode(node.ctx, node)
         else:
             if _is_any_typing_member(node.value, self.scopeStack):
                 with self._enter_annotation():
@@ -1728,7 +1764,10 @@ class Checker(object):
         self.handleChildren(node)
 
     def STR(self, node):
-        if self._in_annotation and not self._in_typing_literal:
+        if (
+                self._in_annotation and
+                not self._in_typing_literal and
+                not self._in_typing_annotated):
             fn = functools.partial(
                 self.handleStringAnnotation,
                 node.s,
--- a/pyflakes/checker.py
+++ b/pyflakes/checker.py
@@ -1272,7 +1272,7 @@ class Checker(object):
             yield
         finally:
             self._in_annotation = orig
-    
+
     @contextlib.contextmanager
     def _enter_typing_annotated(self):
         orig, self._in_typing_annotated = self._in_typing_annotated, True
@@ -1493,8 +1493,9 @@ class Checker(object):
             self.handleNode(node.value, node)
 
             if (
-                isinstance(node.slice, ast.Index) and
-                isinstance(node.slice.value, ast.Tuple)):
+                    isinstance(node.slice, ast.Index) and
+                    isinstance(node.slice.value, ast.Tuple)
+            ):
                 slice_children = list(iter_child_nodes(node.slice.value))
                 if slice_children:
                     self.handleNode(slice_children[0], node.slice)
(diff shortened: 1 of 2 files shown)


sk-:
First pass at fixing #574, this is my first contribution to `pyflakes` so let me know if any changes are required.

Note also that the failing test in Appveyor is because it failed to install `pypy`.

```
(New-Object Net.WebClient).DownloadFile('https://bitbucket.org/pypy/pypy/downloads/pypy2.7-v7.1.1-win32.zip', "$env:appveyor_build_folder\pypy2.7-v7.1.1-win32.zip")
Exception calling "DownloadFile" with "2" argument(s): "The remote server returned an error: (404) Not Found."
At line:1 char:1
+ (New-Object Net.WebClient).DownloadFile('https://bitbucket.org/pypy/p ...
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (:) [], MethodInvocationException
    + FullyQualifiedErrorId : WebException
 
Command executed with exception: Exception calling "DownloadFile" with "2" argument(s): "The remote server returned an error: (404) Not Found."
```

pyflakes/checker.py
```diff
@@ -1471,6 +1480,34 @@ def SUBSCRIPT(self, node):
...
+        elif (
+                (
+                    isinstance(node.value, ast.Name) and
+                    node.value.id == 'Annotated'
+                ) or (
+                    isinstance(node.value, ast.Attribute) and
+                    node.value.attr == 'Annotated'
+                )
+        ):
+            orig = self._in_typing_annotated
```
asottile, line 1492: instead of a new attribute, I think you can visit the annotation part with `_in_annotation` True and visit the non-annotation part with it `False`

pyflakes/test/test_type_annotations.py
```diff
@@ -507,6 +507,51 @@ def f(x: Literal['some string']) -> None:
...
+        """)
+
+    @skipIf(version_info < (3,), 'new in Python 3')
+    def test_forward_ref_missing(self):
+        self.flakes("""
+        from typing import Optional
+
+        def f(x: Optional['integer']) -> None:
+            return None
+        """, m.UndefinedName)
```
asottile, line 526: these two tests are already tested above, I don't think this adds any coverage that isn't already there

pyflakes/test/test_type_annotations.py
```diff
@@ -507,6 +507,51 @@ def f(x: Literal['some string']) -> None:
...
+
+    @skipIf(version_info < (3, 6), 'new in Python 3.6')
+    def test_annotated_type_typing_with_string_args(self):
+        self.flakes("""
+        from typing import Annotated
+
+        def f(x: Annotated[int, '> 0']) -> None:
+            return None
+        """)
```
asottile, line 553: the decorators on these three tests are incorrect -- there's nothing python3.6-specific about these particular tests

asottile:
if you rebase, appveyor should be fixed

pyflakes/test/test_type_annotations.py
sk-: Thanks, I had recalled wrongly that this was supported only for python 3.6 onwards. But it is supported for python >= 3.4. See https://github.com/python/typing/pull/637

pyflakes/checker.py
sk-: Done

pyflakes/test/test_type_annotations.py
sk-: Thanks, I removed those tests.

pyflakes/checker.py
```diff
@@ -1501,9 +1493,9 @@ def SUBSCRIPT(self, node):
                     self.handleNode(slice_children[0], node.slice)
                 # Only the first argument of an Annotated type will always
                 # correspond to an actual type.
-                with self._enter_typing_annotated():
-                    for slice_child in slice_children[1:]:
-                        self.handleNode(slice_child, node)
+                self._in_annotation = False
```
asottile, line 1496: I believe this needs to be restored afterwards (as in a context manager that does the opposite of the one which enables `_in_annotation`)

for example this should fail (and may not with this change?):

```python
def f(x: Union[Annotated['int', '>0'], 'integer']) -> ...:
```

sk-: You are right, that test case failed. I added it as a test case and fixed the code.

asottile (approved):

sk-:
Thanks @asottile for the quick review.

sigmavirus24 (approved):

asottile:
this has a conflict with #535 -- ideally I'd like to merge the other one first (since it was written first) and then update this PR -- thoughts?

sigmavirus24:
@asottile that makes sense. The other one was already hit with a different merge conflict. Have you been holding off on merging that one for a reason?

asottile:
@sigmavirus24 iirc pyflakes has 2-signoff so I was going to wait for a second review (though, given inactivity maybe we should go to 1-signoff?)

sigmavirus24:
> iirc pyflakes has 2-signoff

If it does, it's not configured that way with the settings. I don't remember the requirement being 2 approvals, but I'll go review that one and see if I can approve it.

pyflakes/checker.py
```diff
@@ -1307,12 +1307,19 @@ def _enter_annotation(self, ann_type=AnnotationState.BARE):
 
     @contextlib.contextmanager
     def _exit_annotation(self):
-        orig, self._in_annotation = self._in_annotation, False
+        orig, self._in_annotation = self._in_annotation, AnnotationState.NONE
```
asottile, line 1310: this can reuse `_enter_annotation(ann_type=AnnotationState.NONE)`

pyflakes/checker.py
```diff
@@ -1511,6 +1519,33 @@ def SUBSCRIPT(self, node):
...
+                    isinstance(node.value, ast.Attribute) and
+                    node.value.attr == 'Annotated'
+                )
+        ):
+            self.handleNode(node.value, node)
+
+            if (
+                    isinstance(node.slice, ast.Index) and
+                    isinstance(node.slice.value, ast.Tuple)
+            ):
```
asottile, line 1536: this if block ends up being not-quite-correct in python3.9+  -- the `ast.Index` type is gone and this is just a `Tuple` directly (this leads to one test failure on python3.9 -- let me know if you want me to try and hack on this for ya since it's a bit tricky)

sk-: Thanks @asottile for the heads up. If you could work on this that'd be great.

pyflakes/checker.py
sk-: Done.

asottile:
@sk-  let me know how that looks!

sk-:
@asottile Looks great, specially the extraction of the function `_is_name_or_attr`.

PR #585, opened 2020-10-02
Fix test_invalidEscape with recent PyPy3 versions, closes #584

### LINKED ISSUE #584: test_invalidEscape fails with PyPy3
Hi,
I see that https://github.com/PyCQA/pyflakes/pull/354 re-enabled this test for PyPy, however it fails for me with PyPy 7.3.2 (both the 3.6 and 3.7 variants).
```
FAIL: test_invalidEscape (test.test_api.CheckTests)
----------------------------------------------------------------------
  File "/tmp/pyflakes/pyflakes/test/test_api.py", line 600, in test_invalidEscape
    sourcePath, [decoding_error])
  File "/tmp/pyflakes/pyflakes/test/test_api.py", line 380, in assertHasErrors
    (count, err.getvalue()), (len(errorList), ''.join(errorList)))
AssertionError: Tuples differ: (1, "[14 chars]3l:1:5: (unicode error) 'unicodeescape' codec [79 chars]^\n") != (1, "[14 chars]3l:1:6: (unicode error) 'unicodeescape' codec [80 chars]^\n")

First differing element 1:
"/tmp/tmp_4l9f33l:1:5: (unicode error) 'unicodeescape' codec [78 chars] ^\n"
"/tmp/tmp_4l9f33l:1:6: (unicode error) 'unicodeescape' codec [79 chars] ^\n"

  (1,
-  "/tmp/tmp_4l9f33l:1:5: (unicode error) 'unicodeescape' codec can't decode "
?                      ^

+  "/tmp/tmp_4l9f33l:1:6: (unicode error) 'unicodeescape' codec can't decode "
?                      ^

   'bytes in position 0-1: truncated \\xXX escape\n'
   "foo = '\\xyz'\n"
-  '    ^\n')
+  '     ^\n')
?   +


----------------------------------------------------------------------
```

sbraz:
@asottile This was surprisingly easy 🙂

I'm not sure if we need to account for older PyPy versions, please let me know.

### DIFF
--- a/pyflakes/test/test_api.py
+++ b/pyflakes/test/test_api.py
@@ -515,7 +515,7 @@ def foo(bar=baz, bax):
 """
         with self.makeTempFile(source) as sourcePath:
             if ERROR_HAS_LAST_LINE:
-                if PYPY and sys.version_info >= (3,):
+                if PYPY:
                     column = 7
                 elif sys.version_info >= (3, 8):
                     column = 9
@@ -543,7 +543,7 @@ foo(bar=baz, bax)
 """
         with self.makeTempFile(source) as sourcePath:
             if ERROR_HAS_LAST_LINE:
-                if PYPY and sys.version_info >= (3,):
+                if PYPY:
                     column = 12
                 elif sys.version_info >= (3, 8):
                     column = 14
@@ -578,7 +578,7 @@ foo(bar=baz, bax)
             else:
                 position_end = 1
                 if PYPY:
-                    column = 6
+                    column = 5
                 else:
                     column = 7
                     # Column has been "fixed" since 3.2.4 and 3.3.1
@@ -717,13 +717,6 @@ class IntegrationTests(TestCase):
     """
     Tests of the pyflakes script that actually spawn the script.
     """
-
-    # https://bitbucket.org/pypy/pypy/issues/3069/pypy36-on-windows-incorrect-line-separator
-    if PYPY and sys.version_info >= (3,) and WIN:
-        LINESEP = '\n'
-    else:
-        LINESEP = os.linesep
-
     def setUp(self):
         self.tempdir = tempfile.mkdtemp()
         self.tempfilepath = os.path.join(self.tempdir, 'temp')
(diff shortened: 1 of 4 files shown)


asottile:
probably not, could you also update the pypy versions in CI? I believe those are in .appveyor.yml and .travis.yml

sbraz:
Looks like the newer PyPy version changes the way some tests behave with PyPy2. I'm having trouble reading the AppVeyor logs, I am going to use a matrix to make it look more like Travis.

asottile:
the problem with a matrixed approach there is it significantly increases the runtime :(

we should probably just move to github actions if we approach this as a matrix -- would improve the speed as well

asottile:
I can help with that if you want, let me know!

sbraz:
Is it that bad? Travis looked like it was the slowest to run. I'm still pushing this out of curiosity and we can skip this commit later. I haven't used GH actions yet so your help is welcome.

sbraz:
5 minutes vs 2 minutes, it is slower indeed. Also it doesn't work for PyPy2 for some reason. The PyPy3 failure seems to have to do with `\r\n`.

asottile:
1m51s vs 5m7s -- yeah it's quite a bit slower to matricize appveyor unfortunately -- but if it helps you debug we can leave it like that and put it before finishing

sbraz:
Let's hope it passes this time, I have a good feeling about this!

pyflakes/test/test_api.py
```diff
@@ -718,11 +718,7 @@ class IntegrationTests(TestCase):
     Tests of the pyflakes script that actually spawn the script.
     """
 
-    # https://bitbucket.org/pypy/pypy/issues/3069/pypy36-on-windows-incorrect-line-separator
-    if PYPY and sys.version_info >= (3,) and WIN:
-        LINESEP = '\n'
-    else:
-        LINESEP = os.linesep
+    LINESEP = os.linesep
```
asottile, line 721: oh nice! forgot about this -- you can probably inline `os.linesep` now -- no need to have the class variable

sbraz: Done, I'll be back later tonight or tomorrow. Feel free to edit my commits before merging if you need to chane anything.

asottile (approved):

PR #586, opened 2020-10-02
Fix tests with Python 3.9

sbraz:
Let's see if this passes on Travis's nightly Python build. If not, I will probably add 3.9rc2 to the matrix.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,12 +12,7 @@ matrix:
       dist: xenial
     - python: 3.8
       dist: xenial
-    - python: nightly
-      dist: xenial
-# TODO: https://bugs.python.org/issue40334
-# the PEG parser is currently broken in some cases
-  allow_failures:
-    - python: nightly
+    - python: 3.9-dev
       dist: xenial
 install: pip install tox
 script: tox -e py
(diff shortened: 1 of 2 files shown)


asottile (approved):

asottile:
actually, one other thing -- can you remove these lines? https://github.com/PyCQA/pyflakes/blob/684edfc8833bdb90b0a8ae562fd24d0e1e2fb686/.travis.yml#L17-L21

sbraz:
Done, and it passes!

.travis.yml
```diff
@@ -14,10 +14,5 @@ matrix:
       dist: xenial
     - python: nightly
       dist: xenial
-# TODO: https://bugs.python.org/issue40334
-# the PEG parser is currently broken in some cases
-  allow_failures:
-    - python: nightly
-      dist: xenial
```
asottile, line 21: oh oops now that I look at this, this is 3.10a0 -- should we select `3.9` here explicitly?

sbraz:
@asottile do you think you will release a new version soon? If not, I'll simply backport my patches to Gentoo for the time being.

asottile:
> 
> 
> @asottile do you think you will release a new version soon? If not, I'll simply backport my patches to Gentoo for the time being.

probably not for a while, probably best to backport the patches for now 👍

.travis.yml
sbraz: I've set it to `3.9-dev`. It seems to be the best way at the moment.

PR #588, opened 2020-10-12
Recognize __qualname__ when used in class scope

pilou-:
`pyflakes` doesn't recognize [`__qualname__`](https://docs.python.org/3/glossary.html#term-qualified-name).

Reproducer:
```
$ echo -e "class Foo:\n    __qualname__" | python3
$ pyflakes --version
2.2.0 Python 3.8.6 on Linux
$ echo -e "class Foo:\n    __qualname__" | pyflakes 
<stdin>:2:5 undefined name '__qualname__'
```

Initially reported [here](https://gitlab.com/pycqa/flake8/-/issues/227), relates #259.

### DIFF
--- a/pyflakes/checker.py
+++ b/pyflakes/checker.py
@@ -669,6 +669,10 @@ class DummyNode(object):
         self.col_offset = col_offset
 
 
+class DetectClassScopedMagic:
+    names = dir()
+
+
 # Globally defined names which are not attributes of the builtins module, or
 # are only present on some platforms.
 _MAGIC_GLOBALS = ['__file__', '__builtins__', 'WindowsError']
@@ -1231,7 +1235,7 @@ class Checker(object):
             # the special name __path__ is valid only in packages
             return
 
-        if name == '__module__' and isinstance(self.scope, ClassScope):
+        if name in DetectClassScopedMagic.names and isinstance(self.scope, ClassScope):
             return
 
         # protected with a NameError handler?
--- a/pyflakes/test/test_undefined_names.py
+++ b/pyflakes/test/test_undefined_names.py
@@ -279,6 +279,23 @@ class Test(TestCase):
                 __module__
         ''', m.UndefinedName)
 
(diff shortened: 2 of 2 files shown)


pyflakes/checker.py
```diff
@@ -1234,6 +1234,9 @@ def handleNodeLoad(self, node):
         if name == '__module__' and isinstance(self.scope, ClassScope):
             return
 
+        if not PY2 and name == '__qualname__' and isinstance(self.scope, ClassScope):
```
asottile, line 1237: ~technically this should be (3, 3) but we don't support 3.x (x<4) so this should be fine

pyflakes/checker.py
```diff
@@ -1234,6 +1234,9 @@ def handleNodeLoad(self, node):
...
+            return
+
```
asottile, line 1239: another idea would be to compute this dynamically:

```python
>>> class C:
...     class_scoped_magic = dir()
... 
>>> C.class_scoped_magic
['__module__', '__qualname__']
```

this would be more future-proof -- what do you think?

asottile (commented):
looks good, I'm happy to approve as-is but maybe check out the idea I had :)

pyflakes/checker.py
pilou-: Good idea!

pyflakes/checker.py
```diff
@@ -1231,8 +1231,12 @@ def handleNodeLoad(self, node):
             # the special name __path__ is valid only in packages
             return
 
-        if name == '__module__' and isinstance(self.scope, ClassScope):
-            return
+        if isinstance(self.scope, ClassScope):
+            class C:
+                class_scoped_magic = dir()
```
asottile, line 1236: this should probably be at the module scope in pyflakes so it isn't computed every time

asottile (approved):

sigmavirus24 (approved):
Clever

PR #596, opened 2020-12-03
Test on GitHub Actions

hugovk:
Travis CI has a new pricing model which places limits on open source.

* https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing
* https://www.jeffgeerling.com/blog/2020/travis-cis-new-pricing-plan-threw-wrench-my-open-source-works

Many projects are moving to GitHub Actions instead, which provides testing on Linux, macOS and Windows, with 20 parallel jobs (compared to 5 on Travis and 1 on AppVeyor).

This PR adds testing for modern Python 3.6-3.9 and PyPy3 on the three operating systems. I didn't bother with EOL Pythons here.

Here's an example run:

* https://github.com/hugovk/pyflakes/actions/runs/399197614

Note PyPy3/Windows is excluded for the time being, we're waiting for GHA to include the latest PyPy3 release:

* https://github.com/actions/setup-python/issues/163

I also didn't remove Travis CI or AppVeyor here, but I would suggest doing so.


sigmavirus24 (approved):

asottile:
we should either drop python 2 or add python 2 to this PR and delete the other providers

hugovk:
I've got another branch ready to drop Python 2.7, 3.4 and 3.5.

Shall I merge it in here, or add 2.7 to this, or wait for this to be merged then open a new PR?

hugovk:
> ... or add python 2 to this PR and delete the other providers

Updated to do this, and also added 3.5. (Python 3.4 isn't available on GHA.)

I think it'll be better to have separate PRs, as the other one is quite big. 

It also means the GHA config would already merged into master, which is needed to enable it for this repo.

.github/workflows/test.yml
```diff
@@ -0,0 +1,60 @@
...
+env:
+  FORCE_COLOR: 1
+
+jobs:
+  build:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: ["2.7", "3.5", "3.6", "3.7", "3.8", "3.9", "pypy3"]
```
asottile, line 14: this would need pypy as well

.github/workflows/test.yml
```diff
@@ -0,0 +1,60 @@
...
+        os: [ubuntu-latest, macos-latest, windows-latest]
```
asottile, line 15: I don't think macos grants us any additional coverage and the macos builders are pretty slow and in high demand -- after adding pypy we're looking at 24 builders which seems excessive to me

we can probably get by with:
- test all of them on linux
- test py2 + minimum py3 + maximum py3 + pypy + pypy3 on windows

.github/workflows/test.yml
```diff
@@ -0,0 +1,60 @@
...
+
+      - name: Cache
+        uses: actions/cache@v2
+        with:
+          path: ${{ steps.pip-cache.outputs.dir }}
+          key:
+            ${{ matrix.os }}-${{ matrix.python-version }}-v1-${{
+            hashFiles('**/setup.py') }}
+          restore-keys: |
+            ${{ matrix.os }}-${{ matrix.python-version }}-v1-
```
asottile, line 43: this isn't needed or helpful here -- pyflakes has no dependencies and caching based on the setup.py doesn't gain us anything

.github/workflows/test.yml
```diff
@@ -0,0 +1,60 @@
...
+
+      - name: Install dependencies
+        run: |
+          python -m pip install -U pip
+          python -m pip install -U wheel
+          python -m pip install -U tox
```
asottile, line 49: at the very least, the last two lines can be combined -- though I don't think it's useful to upgrade pip or wheel anyway so this can probabl just be `pip install wheel`

.github/workflows/test.yml
```diff
@@ -0,0 +1,60 @@
...
+
+      - name: Tox tests
+        shell: bash
+        run: |
+          tox -e py
```
asottile, line 54:
```suggestion
        run: tox -e py
```

hugovk:
Ha, AppVeyor fails with no `.appveyor.yml`. It'll need disabling, or at least this option ticking at https://ci.appveyor.com/project/sigmavirus24/pyflakes/settings:

image

.github/workflows/test.yml
```diff
@@ -0,0 +1,60 @@
...
+
+      - name: Upload coverage
+        uses: codecov/codecov-action@v1
```
asottile, line 57: we're not using `codecov`

.github/workflows/test.yml
```diff
@@ -0,0 +1,60 @@
+name: Test
+
+on: [push, pull_request]
+
+env:
+  FORCE_COLOR: 1
```
asottile, line 6: does anything use this?  this isn't a standard environment variable I'm familiar with

hugovk: At least pytest does, I normally just include it as a default. But it's not being passed via tox and the test output is monochrome, so I'll just remove it.

.github/workflows/test.yml
hugovk: Thanks for the review!

Note that pypy is broken on GHA until they include the latest pypy release (see https://github.com/actions/setup-python/issues/163), so pypy2/3 are excluded for now. I've included a comment explaining why.

.github/workflows/test.yml
```diff
@@ -0,0 +1,37 @@
...
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.python-version }}
+
+      - name: Install dependencies
+        run: python -m pip install -U tox
+
+      - name: Tox tests
+        shell: bash
```
asottile, line 36: this line can be removed too

asottile (approved):

sigmavirus24 (approved):

PR #597, opened 2020-12-04
Drop support for EOL Python 2.7, 3.4 and 3.5

hugovk:


pyflakes/api.py
```diff
@@ -187,8 +186,7 @@ def _get_version():
     """
     Retrieve and format package version along with python version & OS used
     """
-    return ('%s Python %s on %s' %
-            (__version__, platform.python_version(), platform.system()))
+    return (f'{__version__} Python {platform.python_version()} on {platform.system()}')
```
asottile, line 189: the parens do nothing here

pyflakes/api.py
```diff
@@ -50,7 +49,7 @@ def check(codeString, filename, reporter=None):
                 lines = codeString.splitlines()
                 if len(lines) >= lineno:
                     text = lines[lineno - 1]
-                    if sys.version_info >= (3, ) and isinstance(text, bytes):
+                    if isinstance(text, bytes):
```
asottile, line 52: this should be impossible I believe, this api should always take text or always take ytes

pyflakes/checker.py
```diff
@@ -19,86 +19,46 @@
...
-PY36_PLUS = sys.version_info >= (3, 6)    # Python 3.6 and above
 PY38_PLUS = sys.version_info >= (3, 8)
 try:
     sys.pypy_version_info
     PYPY = True
 except AttributeError:
     PYPY = False
 
-builtin_vars = dir(__import__('__builtin__' if PY2 else 'builtins'))
+builtin_vars = dir(__import__('builtins'))
```
asottile, line 29: this should `import builtins` at the module scope

pyflakes/checker.py
```diff
@@ -1546,7 +1493,7 @@ def SUBSCRIPT(self, node):
 
     def _handle_string_dot_format(self, node):
         try:
-            placeholders = tuple(parse_format_string(node.func.value.s))
+            placeholders = tuple(string.Formatter().parse(node.func.value.s))
```
asottile, line 1496: don't inline this, it's intentionally module scoped because it is expensive to construct a `Formatter()` repeatedly

pyflakes/checker.py
```diff
@@ -19,86 +19,46 @@
...
 
-if PY2:
-    def getNodeType(node_class):
-        # workaround str.upper() which is locale-dependent
-        return str(unicode(node_class.__name__).upper())
 
-    def get_raise_argument(node):
-        return node.type
+def get_raise_argument(node):
+    return node.exc
```
asottile, line 37: this should be inlined

pyflakes/checker.py
```diff
@@ -19,86 +19,46 @@
...
+    if isinstance(n, ast.If):
+        return [n.body]
+    if isinstance(n, ast.Try):
+        return [n.body + n.orelse] + [[hdl] for hdl in n.handlers]
+
+
+FOR_TYPES = (ast.For, ast.AsyncFor)
+LOOP_TYPES = (ast.While, ast.For, ast.AsyncFor)
+FUNCTION_TYPES = (ast.FunctionDef, ast.AsyncFunctionDef)
+ANNASSIGN_TYPES = (ast.AnnAssign,)
```
asottile, line 50: these can all be inlined

pyflakes/checker.py
```diff
@@ -112,9 +72,7 @@ def _is_tuple_constant(node):  # type: (ast.AST) -> bool
     def _is_constant(node):
         return isinstance(node, ast.Constant) or _is_tuple_constant(node)
 else:
-    _const_tps = (ast.Str, ast.Num)
-    if not PY2:
-        _const_tps += (ast.Bytes,)
+    _const_tps = (ast.Str, ast.Num, ast.Bytes)
```
asottile, line 75: this can be inlined

pyflakes/checker.py
```diff
@@ -675,10 +629,8 @@ class DetectClassScopedMagic:
 
 # Globally defined names which are not attributes of the builtins module, or
 # are only present on some platforms.
-_MAGIC_GLOBALS = ['__file__', '__builtins__', 'WindowsError']
 # module scope annotation will store in `__annotations__`, see also PEP 526.
-if PY36_PLUS:
-    _MAGIC_GLOBALS.append('__annotations__')
+_MAGIC_GLOBALS = ['__file__', '__builtins__', 'WindowsError', '__annotations__']
```
asottile, line 633: this can be a tuple now

pyflakes/checker.py
```diff
@@ -859,8 +811,7 @@ class Checker(object):
         ast.GeneratorExp: GeneratorScope,
         ast.DictComp: GeneratorScope,
     }
-    if PY35_PLUS:
-        _ast_node_scope[ast.AsyncFunctionDef] = FunctionScope
+    _ast_node_scope[ast.AsyncFunctionDef] = FunctionScope
```
asottile, line 814: this can be statically part of the literal above

pyflakes/checker.py
```diff
@@ -1945,7 +1892,7 @@ def GENERATOREXP(self, node):
         self.handleChildren(node)
         self.popScope()
 
-    LISTCOMP = handleChildren if PY2 else GENERATOREXP
+    LISTCOMP = GENERATOREXP
```
asottile, line 1895: this can be part of the line below

pyflakes/checker.py
```diff
@@ -2036,24 +1981,14 @@ def LAMBDA(self, node):
...
                 args.append(arg.arg)
                 annotations.append(arg.annotation)
-            defaults = node.args.defaults + node.args.kw_defaults
+        for arg in node.args.args + node.args.kwonlyargs:
+            args.append(arg.arg)
+            annotations.append(arg.annotation)
+        defaults = node.args.defaults + node.args.kw_defaults
 
         # Only for Python3 FunctionDefs
         is_py3_func = hasattr(node, 'returns')
```
asottile, line 1994: this should always be true

pyflakes/checker.py
```diff
@@ -2062,13 +1997,9 @@ def addArgs(arglist):
             wildcard = getattr(node.args, arg_name)
```
asottile, line 1997: this shouldn't need getattr any more

pyflakes/test/test_api.py
```diff
@@ -635,11 +610,10 @@ def test_encodedFileUTF8(self):
         """
         If source file declares the correct encoding, no error is reported.
         """
-        SNOWMAN = unichr(0x2603)
-        source = ("""\
-# coding: utf-8
-x = "%s"
-""" % SNOWMAN).encode('utf-8')
+        SNOWMAN = chr(0x2603)
```
asottile, line 613: we're not fighting with py2 / py3.2 any more, it would be more readable to put the actual character in this string

pyflakes/test/test_api.py
```diff
@@ -655,23 +629,12 @@ def test_misencodedFileUTF8(self):
...
         """
-        SNOWMAN = unichr(0x2603)
-        source = ("""\
-# coding: ascii
-x = "%s"
-""" % SNOWMAN).encode('utf-8')
+        SNOWMAN = chr(0x2603)
+        source = f"""# coding: ascii
+x = "{SNOWMAN}\"
+""".encode('utf-8')
```
asottile, line 635: .encode('utf-8') => .encode()

setup.cfg
```diff
@@ -1,5 +1,2 @@
-[bdist_wheel]
-universal = 1
```
asottile, line 2: technically this is still valid and should be kept -- a universal wheel contains no version-specific code and is pure python (which we still qualify for)

hugovk: It does give a bit of extra protection by naming the file `pyflakes-2.2.0-py3-none-any.whl` instead of `pyflakes-2.2.0-py2.py3-none-any.whl`, and then pip on Python 2 gives this error:

```
ERROR: pyflakes-2.2.0-py3-none-any.whl is not a supported wheel on this platform.
```

And that's before it gets to the `python_requires` check. I think the wheel check has been around longer than the `python_requires` check?

Anyway, I've re-instated this.

---

The `python_requires` check for `pyflakes-2.2.0-py2.py3-none-any.whl` looks like:

```
ERROR: Package 'pyflakes' requires a different Python: 2.7.17 not in '>=3.6'
```

pyflakes/checker.py
hugovk: When I tried https://github.com/hugovk/pyflakes/commit/79ff7c04aa2d8432d2ce8c8d9e7c48ff3124c8af, it failed:

```
  File "/home/runner/work/pyflakes/pyflakes/pyflakes/checker.py", line 1994, in LAMBDA
    annotations.append(node.returns)
AttributeError: 'Lambda' object has no attribute 'returns'
```

https://github.com/hugovk/pyflakes/actions/runs/403118601

pyflakes/checker.py
hugovk: Please could you elaborate on how to change it?

pyflakes/checker.py
asottile: hmmm perhaps `LAMBDA` should go through a different code path 🤔

pyflakes/checker.py
asottile: should be able to loop over `(node.varag, node.kwarg)` directly (or whatever the names are)

hugovk: https://github.com/hugovk/pyflakes/commit/02728fcf9b7dee33d9914ab09a05f8d7d4f9c54e fails like https://github.com/hugovk/pyflakes/actions/runs/404149955:

```
  File "/home/runner/work/pyflakes/pyflakes/pyflakes/checker.py", line 842, in __init__
    self.handleChildren(tree)
  File "/home/runner/work/pyflakes/pyflakes/pyflakes/checker.py", line 1288, in handleChildren
    self.handleNode(node, tree)
  File "/home/runner/work/pyflakes/pyflakes/pyflakes/checker.py", line 1335, in handleNode
    handler(node)
  File "/home/runner/work/pyflakes/pyflakes/pyflakes/checker.py", line 1957, in FUNCTIONDEF
    self.LAMBDA(node)
  File "/home/runner/work/pyflakes/pyflakes/pyflakes/checker.py", line 1984, in LAMBDA
    for wildcard in (node.vararg, node.kwarg):
AttributeError: 'FunctionDef' object has no attribute 'vararg'
```

asottile: then I think this code is dead, those were the python2 attributes!

hugovk: Hmm, removing it like https://github.com/hugovk/pyflakes/commit/5a149a93b3581076ebd0a9490ab4cb141de2f641 fails like https://github.com/hugovk/pyflakes/actions/runs/404358215:

```
FAIL: test_annotationUndefined (test.test_doctests.TestUndefinedNames)
Undefined annotations.
----------------------------------------------------------------------
  File "/home/runner/work/pyflakes/pyflakes/pyflakes/test/test_undefined_names.py", line 657, in test_annotationUndefined
    ''')
  File "/home/runner/work/pyflakes/pyflakes/pyflakes/test/test_doctests.py", line 51, in flakes
    return super().flakes(self.doctestify(input), *args, **kw)
  File "/home/runner/work/pyflakes/pyflakes/pyflakes/test/harness.py", line 36, in flakes
    {}'''.format(input, expectedOutputs, '\n'.join([str(o) for o in w.messages])))
AssertionError: Lists differ: [<class 'pyflakes.messages.UnusedImport'>,[37 chars]rt'>] != []

First list contains 2 additional elements.
First extra element 0:
<class 'pyflakes.messages.UnusedImport'>

+ []
- [<class 'pyflakes.messages.UnusedImport'>,
-  <class 'pyflakes.messages.UnusedImport'>] : for input:
def doctest_something():
    """
       
       >>> from abc import note1, note2, note3, note4, note5
       >>> def func(a: note1, *args: note2,
       ...          b: note3=12, **kw: note4) -> note5: pass
    """

expected outputs:
[]
but got:
(none):4:12 'abc.note2' imported but unused
(none):4:12 'abc.note4' imported but unused
```

asottile: oh right, it's `(node.args.vararg, node.args.kwarg)` not `(node.vararg, node.kwarg)`

hugovk: Done, thanks!

pyflakes/checker.py
```diff
@@ -2036,39 +1969,24 @@ def LAMBDA(self, node):
...
                 args.append(arg.arg)
                 annotations.append(arg.annotation)
-            defaults = node.args.defaults + node.args.kw_defaults
+        for arg in node.args.args + node.args.kwonlyargs:
+            args.append(arg.arg)
+            annotations.append(arg.annotation)
+        defaults = node.args.defaults + node.args.kw_defaults
 
         # Only for Python3 FunctionDefs
         is_py3_func = hasattr(node, 'returns')
```
asottile, line 1982: This should maybe be named `is_non_lambda`

pyflakes/test/test_api.py
```diff
@@ -495,11 +487,10 @@ def test_eofSyntaxErrorWithTab(self):
 
             self.assertHasErrors(
                 sourcePath,
-                ["""\
```
asottile, line 498: I prefer the escaped newline so the full error message is all on the left -- not a big deal though

pyflakes/test/test_api.py
```diff
@@ -635,11 +610,9 @@ def test_encodedFileUTF8(self):
         """
         If source file declares the correct encoding, no error is reported.
         """
-        SNOWMAN = unichr(0x2603)
-        source = ("""\
-# coding: utf-8
-x = "%s"
-""" % SNOWMAN).encode('utf-8')
+        source = """# coding: utf-8
+x = "☃\"
```
asottile, line 614: stray backslash here?

pyflakes/test/test_api.py
```diff
@@ -681,14 +642,12 @@ def test_misencodedFileUTF16(self):
         If a source file contains bytes which cannot be decoded, this is
         reported on stderr.
         """
-        SNOWMAN = unichr(0x2603)
-        source = ("""\
-# coding: ascii
-x = "%s"
-""" % SNOWMAN).encode('utf-16')
+        source = """# coding: ascii
+x = "☃\"
```
asottile, line 646: Also here

asottile (approved):
I don't really have the final say here on the dropping of python 2 -- but I'm in favor of this patch -- just a few more small things to fix up I think

setup.cfg
sigmavirus24: Looking at https://packaging.python.org/guides/distributing-packages-using-setuptools/#pure-python-wheels it does appear that we should be removing `universal` because universal is for py2 & py3 otherwise just pure-python wheels

pyflakes/api.py
```diff
@@ -50,11 +49,10 @@ def check(codeString, filename, reporter=None):
                 lines = codeString.splitlines()
                 if len(lines) >= lineno:
                     text = lines[lineno - 1]
-                    if sys.version_info >= (3, ) and isinstance(text, bytes):
-                        try:
-                            text = text.decode('ascii')
-                        except UnicodeDecodeError:
-                            text = None
+                    try:
+                        text = text.decode('ascii')
```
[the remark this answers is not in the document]
hugovk: https://github.com/hugovk/pyflakes/commit/9c4352b5b77b63c4a9d9525b11458afd31a88e22 fails for PyPy3: https://github.com/hugovk/pyflakes/actions/runs/406812006

hugovk:
Thanks for the review! Updated!

setup.cfg
sigmavirus24: @asottile `universal` wheels according to the official packaging docs (see last comment) are meant to advertise support for Py2 and Py3 both. At this point, the package won't actually work on Py2 and advertising support for it is incorrect. Do you have information from the PyPA contradicting their documentation?

hugovk: Comparing wheel filenames created with and without the flag:

```
pyflakes-2.2.0-py2.py3-none-any.whl
pyflakes-2.2.0-py3-none-any.whl
```

Their contents are identical, except for `pyflakes-2.2.0.dist-info/WHEEL` (and `RECORD`'s hash of this file):

```
Wheel-Version: 1.0
Generator: bdist_wheel (0.36.0)
Root-Is-Purelib: true
Tag: py2-none-any
Tag: py3-none-any
```
```
Wheel-Version: 1.0
Generator: bdist_wheel (0.36.0)
Root-Is-Purelib: true
Tag: py3-none-any
```

asottile: perhaps the nuance of that field has changed over time -- my thought is that for python 4 it would still be correct for pure python packages to claim "universal" status

sigmavirus24: Wait, we're already on Python 4? :stuck_out_tongue:

hugovk: At least right now universal only means 2 + 3.

Shall I remove `universal = 1`?

hugovk:
(That quick merge from `master` didn't work, will undo and redo it properly!)

PR #600, opened 2020-12-25
fix: support type annotation of recursive data structure in local scope

### LINKED ISSUE #567: False positive in type annotation of recursive data structure.
```python
from typing import Dict

def test() -> None:
    class Tree(Dict[str, "Tree"]):
        def __missing__(self, key: str) -> "Tree":
            self[key] = Tree()
            return self[key]
```
```
/tmp/x.py:4:26 undefined name 'Tree'
```

PrettyWood:
Hi everyone !
This is my first contribution so the whole code is still pretty new to me.
While working on _pydantic_, I encountered a [false positive](https://github.com/samuelcolvin/pydantic/pull/2221/commits/ab57d19cd77230cba041ee855b2a93d03ad40614) and saw there was already an opened issue.
I hence decided to give it a go!

closes #567

### DIFF
--- a/pyflakes/test/test_type_annotations.py
+++ b/pyflakes/test/test_type_annotations.py
@@ -667,3 +667,22 @@ class TestTypeAnnotations(TestCase):
                 def f():  # type: () -> int
                     pass
         """)
+
+    @skipIf(version_info < (3, 6), 'new in Python 3.6')
+    def test_recursive_global_scope(self):
+        self.flakes("""
+            from typing import Optional, Tuple
+
+            class NestedTuple:
+                x: Tuple[int, Optional['NestedTuple']]
+        """)
+
+    @skipIf(version_info < (3, 6), 'new in Python 3.6')
+    def test_recursive_local_scope(self):
+        self.flakes("""
+            from typing import Optional, Tuple
+
+            def test():
+                class NestedTuple:
+                    x: Tuple[int, Optional['NestedTuple']]
+        """)
--- a/pyflakes/checker.py
+++ b/pyflakes/checker.py
@@ -866,7 +866,6 @@ class Checker(object):
     offset = None
     traceTree = False
     _in_annotation = AnnotationState.NONE
-    _in_deferred = False
 
     builtIns = set(builtin_vars).union(_MAGIC_GLOBALS)
     _customBuiltIns = os.environ.get('PYFLAKES_BUILTINS')
@@ -898,7 +897,6 @@ class Checker(object):
         for builtin in self.builtIns:
             self.addBinding(None, Builtin(builtin))
         self.handleChildren(tree)
-        self._in_deferred = True
         self.runDeferred(self._deferredFunctions)
         # Set _deferredFunctions to None so that deferFunction will fail
         # noisily if called after we've run through the deferred functions.
@@ -1803,10 +1801,7 @@ class Checker(object):
                 node.col_offset,
                 messages.ForwardAnnotationSyntaxError,
             )
-            if self._in_deferred:
-                fn()
-            else:
-                self.deferFunction(fn)
+            self.deferFunction(fn)
 
     if PY38_PLUS:
         def CONSTANT(self, node):


pyflakes/test/test_type_annotations.py
```diff
@@ -667,3 +667,20 @@ class C(Protocol):
                 def f():  # type: () -> int
                     pass
         """)
+
+    def test_recursive_global(self):
```
PrettyWood, line 671: This was already working properly

pyflakes/test/test_type_annotations.py
```diff
@@ -667,3 +667,20 @@ class C(Protocol):
...
+        self.flakes("""
+            from typing import Optional, Tuple
+
+            class NestedTuple:
+                x: Tuple[int, Optional['NestedTuple']]
+        """)
+
+    def test_recursive_local(self):
```
PrettyWood, line 679: This used to fail because on the first check the ast just has the "function def" node
So when checking the inside, `self._in_deferred` was now set to `True` causing the issue.
I don't really know why we can't just run everything without it. I remove the  `_in_deferred` boolean and all the unit tests work. I also checked locally on different projects without any issue.
If this causes other problems, I'll be happy to add an extra unit test for this and work on another way to fix it

pyflakes/test/test_type_annotations.py
```diff
@@ -667,3 +667,22 @@ class C(Protocol):
                 def f():  # type: () -> int
                     pass
         """)
+
+    @skipIf(version_info < (3, 6), 'new in Python 3.6')
+    def test_recursive_global_scope(self):
```
PrettyWood, line 672: This was already working properly

pyflakes/test/test_type_annotations.py
```diff
@@ -667,3 +667,22 @@ class C(Protocol):
...
+        self.flakes("""
+            from typing import Optional, Tuple
+
+            class NestedTuple:
+                x: Tuple[int, Optional['NestedTuple']]
+        """)
+
+    @skipIf(version_info < (3, 6), 'new in Python 3.6')
+    def test_recursive_local_scope(self):
```
PrettyWood, line 681: This used to fail because on the first check the ast just has the "function def" node
So when checking the inside, `self._in_deferred` was now set to `True` causing the issue.
I don't really know why we can't just run everything without it. I remove the  `_in_deferred` boolean and all the unit tests work. I also checked locally on different projects without any issue.
If this causes other problems, I'll be happy to add an extra unit test for this and work on another way to fix it

.gitignore
```diff
@@ -6,3 +6,5 @@
 /dist/
 /build/
 docs/_build
+.idea/
+.python-version
```
asottile, line 10: please don't edit .gitignore files on projects you don't maintain -- you can put these in a personal global gitignore (please rebase this out)

pyflakes/test/test_type_annotations.py
asottile: yeah I can't seem to find why I wrote that either -- perhaps I was afraid of modifying the `deferredFunctions` list while it was being iterated -- but it seems append-only-while-iterating works fine in python

.gitignore
PrettyWood: Alright! It was only a suggestion hence the distinct commit 😉

asottile (approved):
seems fine, I'll throw this at some typed code just to double-make-sure

asottile:
Ok so I ran this against a bunch of typed code, I can't really explain this difference but I boiled this down from a change in `pandas`:

```python
from typing import Union

if TYPE_CHECKING:
    from x import Y

def f():
    x: Union[int, "Y"]

    if f:
        from x import Y
        return Y()
```

Before this produced no errors, but after your PR it produces the following:

```console
$ python3 -m pyflakes t.py
t.py:5:5 'x.Y' imported but unused
t.py:11:9 redefinition of unused 'Y' from line 5
```

asottile:
via #754

PR #604, opened 2021-01-04
deterministic ordering for errors in __all__

### LINKED ISSUE #603: "undefined name in __all__" has non-deterministic ordering
Found this while putting together a tool to look for regressions in the latest release and noticed that this produces things out-of-order!

```console
$ cat t.py 
__all__ = ('a', 'b', 'c', 'd')
$ python3 -m pyflakes t.py
t.py:1:1 undefined name 'a' in __all__
t.py:1:1 undefined name 'd' in __all__
t.py:1:1 undefined name 'b' in __all__
t.py:1:1 undefined name 'c' in __all__
$ python3 -m pyflakes t.py
t.py:1:1 undefined name 'c' in __all__
t.py:1:1 undefined name 'd' in __all__
t.py:1:1 undefined name 'b' in __all__
t.py:1:1 undefined name 'a' in __all__
```

This is due to an intermediate set in the code for that check: https://github.com/PyCQA/pyflakes/blob/0e4194b238ffd006441f1a33373062d9c7272d0e/pyflakes/checker.py#L993-L997

asottile:
Resolves #603

I couldn't figure out a way to test this that wasn't flakey and without writing a bunch of test infra to look at the actual messages (since they're all the same type)

Here's a demo of the fix however:

```console
$ python3 -m pyflakes t.py
t.py:1:1 undefined name 'a' in __all__
t.py:1:1 undefined name 'b' in __all__
t.py:1:1 undefined name 'c' in __all__
t.py:1:1 undefined name 'd' in __all__
$ python3 -m pyflakes t.py
t.py:1:1 undefined name 'a' in __all__
t.py:1:1 undefined name 'b' in __all__
t.py:1:1 undefined name 'c' in __all__
t.py:1:1 undefined name 'd' in __all__
$ python3 -m pyflakes t.py
t.py:1:1 undefined name 'a' in __all__
t.py:1:1 undefined name 'b' in __all__
t.py:1:1 undefined name 'c' in __all__
t.py:1:1 undefined name 'd' in __all__
$ python3 -m pyflakes t.py
t.py:1:1 undefined name 'a' in __all__
t.py:1:1 undefined name 'b' in __all__
t.py:1:1 undefined name 'c' in __all__
t.py:1:1 undefined name 'd' in __all__
$ python3 -m pyflakes t.py
t.py:1:1 undefined name 'a' in __all__
t.py:1:1 undefined name 'b' in __all__
t.py:1:1 undefined name 'c' in __all__
t.py:1:1 undefined name 'd' in __all__
```

### DIFF
--- a/pyflakes/checker.py
+++ b/pyflakes/checker.py
@@ -992,7 +992,10 @@ class Checker(object):
 
             if all_binding:
                 all_names = set(all_binding.names)
-                undefined = all_names.difference(scope)
+                undefined = [
+                    name for name in all_binding.names
+                    if name not in scope
+                ]
             else:
                 all_names = undefined = []
 


pyflakes/checker.py
```diff
@@ -992,7 +992,10 @@ def checkDeadScopes(self):
 
             if all_binding:
                 all_names = set(all_binding.names)
-                undefined = all_names.difference(scope)
+                undefined = [
```
sigmavirus24, line 995: I wonder if we can drop `all_names` or if the usage on L1026 is really that much faster just because it's a set in practice on real code-bases

sigmavirus24 (approved):
Non-blocking thought in-line

pyflakes/checker.py
asottile: Yeah I thought about this, but I don't really have a good way to profile this.  I imagine it's not a significant difference to keep the original value as a list and just do `in` against that (however it does appear to change this from `O(N) + O(M)` to  `O(N * M)` where N is the number of imports and M is the number of items in `__all__`

Probably ok to just leave it since it keeps this change a little bit more minimal

asottile:
@sigmavirus24 while I have you, can you click disable in appveyor so we don't get the empty failure?

PR #606, opened 2021-01-04
Fix annotation clobbering __all__

### LINKED ISSUE #605: Odd edge-case regression in type annotated `__all__`
I'm not quite sure how to approach fixing this -- it regressed as part of #535 -- CC @srittau 

Here's a simplified example, this is boiled down from numpy's `numpy/typing/__init__.py`

```python
from typing import TYPE_CHECKING, List

from y import z

if not TYPE_CHECKING:
    __all__ = ("z",)
else:
    __all__: List[str]
```

After #535, the type annotation clobbers the `__all__` "`ExportBinding`" which causes it to no longer mark the imports as used

I *think* we can make `Annotated` a `setdefault` on the latest namespace to prevent this?

asottile:
Resolves #605

### DIFF
--- a/pyflakes/checker.py
+++ b/pyflakes/checker.py
@@ -1140,7 +1140,10 @@ class Checker(object):
             # then assume the rebound name is used as a global or within a loop
             value.used = self.scope[value.name].used
 
-        self.scope[value.name] = value
+        # don't treat annotations as assignments if there is an existing value
+        # in scope
+        if value.name not in self.scope or not isinstance(value, Annotation):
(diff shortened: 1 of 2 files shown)


sigmavirus24 (approved):

sigmavirus24:
I turned off the app-veyor webhook for pyflakes.

sigmavirus24:
Also went to appveyor and deleted this project

srittau (approved):
Makes sense to me.

PR #609, opened 2021-01-14
Fix: Named types incorrectly raising UndefinedName inside annotated subscript

### LINKED ISSUE #575: Nested TypedDict not supported
Flake8 returns strange `undefined name` errors when using nested [alternative syntax style](https://www.python.org/dev/peps/pep-0589/#alternative-syntax) `TypedDict` declarations.

Example:

```py
from typing import TypedDict


class Example(TypedDict):
    nested: TypedDict("Nested", {"foo/bar": str})
```

Pyflakes returns the following:
```sh
foo.py:5:23 undefined name 'Nested'
foo.py:5:34 undefined name 'foo'
foo.py:5:34 undefined name 'bar'
```

Version information:
```sh
$ pyflakes --version
2.2.0 Python 3.8.0 on Darwin
```

A workaround is to simply not nest the declarations:

```py
from typing import TypedDict


Nested = TypedDict("Nested", {"foo/bar": str})


class Example(TypedDict):
    nested: Nested
```

legau:
Hi,
this PR closes #575

When inside an annotated subscript, named types (NamedTuple and TypedDict) and their string arguments are flagged as a forward reference which should not be the case.

### DIFF
--- a/pyflakes/checker.py
+++ b/pyflakes/checker.py
@@ -1678,28 +1678,79 @@ class Checker(object):
         ):
             self._handle_string_dot_format(node)
 
+        omit = []
+        annotated = []
+        not_annotated = []
+
         if (
             _is_typing(node.func, 'cast', self.scopeStack) and
-            len(node.args) >= 1 and
-            isinstance(node.args[0], ast.Str)
+            len(node.args) >= 1
         ):
-            with self._enter_annotation(AnnotationState.STRING):
+            with self._enter_annotation():
                 self.handleNode(node.args[0], node)
 
         elif _is_typing(node.func, 'TypeVar', self.scopeStack):
+
             # TypeVar("T", "int", "str")
-            for arg in node.args[1:]:
-                if isinstance(arg, ast.Str):
-                    with self._enter_annotation():
-                        self.handleNode(arg, node)
+            omit += ["args"]
+            annotated += [arg for arg in node.args[1:]]
 
             # TypeVar("T", bound="str")
-            for keyword in node.keywords:
-                if keyword.arg == 'bound' and isinstance(keyword.value, ast.Str):
-                    with self._enter_annotation():
-                        self.handleNode(keyword.value, node)
+            omit += ["keywords"]
+            annotated += [k.value for k in node.keywords if k.arg == "bound"]
+            not_annotated += [
+                (k, ["value"] if k.arg == "bound" else None)
+                for k in node.keywords
+            ]
 
-        self.handleChildren(node)
+        elif _is_typing(node.func, "TypedDict", self.scopeStack):
+            # TypedDict("a", {"a": int})
+            if len(node.args) > 1 and isinstance(node.args[1], ast.Dict):
+                omit += ["args"]
+                annotated += node.args[1].values
+                not_annotated += [
+                    (arg, ["values"] if i == 1 else None)
+                    for i, arg in enumerate(node.args)
+                ]
+
+            # TypedDict("a", a=int)
+            omit += ["keywords"]
+            annotated += [k.value for k in node.keywords]
+            not_annotated += [(k, ["value"]) for k in node.keywords]
+
+        elif _is_typing(node.func, "NamedTuple", self.scopeStack):
+            # NamedTuple("a", [("a", int)])
+            if (
+                len(node.args) > 1 and
+                isinstance(node.args[1], (ast.Tuple, ast.List)) and
+                all(isinstance(x, (ast.Tuple, ast.List)) and
+                    len(x.elts) == 2 for x in node.args[1].elts)
+            ):
+                omit += ["args"]
+                annotated += [elt.elts[1] for elt in node.args[1].elts]
+                not_annotated += [(elt.elts[0], None) for elt in node.args[1].elts]
+                not_annotated += [
+                    (arg, ["elts"] if i == 1 else None)
+                    for i, arg in enumerate(node.args)
+                ]
+                not_annotated += [(elt, "elts") for elt in node.args[1].elts]
+
+            # NamedTuple("a", a=int)
+            omit += ["keywords"]
+            annotated += [k.value for k in node.keywords]
+            not_annotated += [(k, ["value"]) for k in node.keywords]
+
+        if omit:
+            with self._enter_annotation(AnnotationState.NONE):
+                for na_node, na_omit in not_annotated:
+                    self.handleChildren(na_node, omit=na_omit)
+                self.handleChildren(node, omit=omit)
+
+            with self._enter_annotation():
+                for annotated_node in annotated:
+                    self.handleNode(annotated_node, node)
+        else:
+            self.handleChildren(node)
 
     def _handle_percent_format(self, node):
         try:
--- a/pyflakes/test/test_type_annotations.py
+++ b/pyflakes/test/test_type_annotations.py
@@ -695,3 +695,44 @@ class TestTypeAnnotations(TestCase):
                 def f():  # type: () -> int
                     pass
         """)
+
+    def test_typednames_correct_forward_ref(self):
+        self.flakes("""
+            from typing import TypedDict, List, NamedTuple
+
+            List[TypedDict("x", {})]
+            List[TypedDict("x", x=int)]
+            List[NamedTuple("a", a=int)]
+            List[NamedTuple("a", [("a", int)])]
+        """)
+        self.flakes("""
+            from typing import TypedDict, List, NamedTuple, TypeVar
+
+            List[TypedDict("x", {"x": "Y"})]
+            List[TypedDict("x", x="Y")]
+            List[NamedTuple("a", [("a", "Y")])]
+            List[NamedTuple("a", a="Y")]
+            List[TypedDict("x", {"x": List["a"]})]
+            List[TypeVar("A", bound="C")]
+            List[TypeVar("A", List["C"])]
+        """, *[m.UndefinedName]*7)
+        self.flakes("""
+            from typing import NamedTuple, TypeVar, cast
+            from t import A, B, C, D, E
+
+            NamedTuple("A", [("a", A["C"])])
+            TypeVar("A", bound=A["B"])
+            TypeVar("A", A["D"])
+            cast(A["E"], [])
+        """)
+
+    @skipIf(version_info < (3, 6), 'new in Python 3.6')
+    def test_namedtypes_classes(self):
+        self.flakes("""
+            from typing import TypedDict, NamedTuple
+            class X(TypedDict):
+                y: TypedDict("z", {"zz":int})
+
+            class Y(NamedTuple):
+                y: NamedTuple("v", [("vv", int)])
+        """)


pyflakes/checker.py
```diff
@@ -1699,7 +1699,12 @@ def CALL(self, node):
                     with self._enter_annotation():
                         self.handleNode(keyword.value, node)
 
-        self.handleChildren(node)
+        if any(_is_typing(node.func, nt, self.scopeStack)
+                for nt in ("TypedDict", "NamedTuple")):
+            with self._enter_annotation(AnnotationState.NONE):
```
asottile, line 1704: this isn't correct, consider `TypedDict("X", {"y": "T"})`

legau: Indeed, my new commit should resolve this

pyflakes/checker.py
```diff
@@ -1699,7 +1699,34 @@ def CALL(self, node):
                     with self._enter_annotation():
                         self.handleNode(keyword.value, node)
 
-        self.handleChildren(node)
+        if any(_is_typing(node.func, nt, self.scopeStack)
+                for nt in ("TypedDict", "NamedTuple")):
```
asottile, line 1703: this `any` no longer buys you anything because you have separate cases below (it already was a little bit of an over-DRYing before -- two things isn't repeating yourself)

pyflakes/checker.py
```diff
@@ -1699,7 +1699,34 @@ def CALL(self, node):
...
+            with self._enter_annotation(AnnotationState.NONE):
+                self.handleChildren(node)
```
asottile, line 1705: this causes the things to be processed more than once which might be problematic -- though I can't think of a case off the top of my head where this produces incorrect results

pyflakes/checker.py
```diff
@@ -1699,7 +1699,34 @@ def CALL(self, node):
...
+
+            attributes = []
+
+            if node.func.id == "TypedDict":
```
asottile, line 1709: TypedDict also supports `TypedDict('foo', bar="T")` form as well

pyflakes/checker.py
```diff
@@ -1699,7 +1699,34 @@ def CALL(self, node):
...
+                if len(node.args) > 1 and isinstance(node.args[1], ast.Dict):
+                    attributes += node.args[1].values
+
+            elif node.func.id == "NamedTuple":
+                # NamedTuple("a", [("a", int)])
+                if len(node.args) > 1 and hasattr(node.args[1], "elts") and all(
```
asottile, line 1715: instead of `hasattr(..., elts)` I would prefer a more explicit `and isinstance(node.args[1], (ast.List, ast.Tuple))` -- same for the inner one as well (instead of `getattr)`

pyflakes/test/test_type_annotations.py
```diff
@@ -695,3 +695,24 @@ class C(Protocol):
...
+    def test_typeddict_allows_any_names(self):
+        self.flakes("""
+            from typing import TypedDict, List, NamedTuple
+
+            List[TypedDict("x", {})]
+            List[NamedTuple("a", a=int)]
+            List[NamedTuple("a", [("a", int)])]
+
+            class X(TypedDict):
+                y: TypedDict("z", {"zz":int})
```
asottile, line 709: this is the only syntax which is python3.6-specific -- the other tests should be tested in all the other python versions as well

pyflakes/checker.py
```diff
@@ -1699,7 +1699,34 @@ def CALL(self, node):
...
+                    len(getattr(x, "elts", [])) == 2 for x in node.args[1].elts
+                ):
+                    attributes = [el.elts[1] for el in node.args[1].elts]
+
+                # NamedTuple("a", a=int)
+                attributes += [k.value for k in node.keywords]
+
+            for a in attributes:
```
asottile, line 1723: I'd prefer a better name than `a`

pyflakes/checker.py
```diff
@@ -1699,7 +1699,34 @@ def CALL(self, node):
...
+                if isinstance(a, ast.Str):
```
asottile, line 1724: I believe this is insufficient, consider `NamedTuple("T", ("a", List["Something"]))`

pyflakes/checker.py
legau: It is in the same vein as the special case TypeVar

pyflakes/checker.py
legau: This is already covered by [this](https://github.com/PyCQA/pyflakes/pull/609/files#diff-5df44d79406311387e0056faae7a8092d1ce9cc85c7d15420186e45a0c712283R1705). See [this test](https://github.com/PyCQA/pyflakes/pull/609/files#diff-a3d1d43d9902e26c2a2893ffbeb0fc7af7d94f35b39f8beaddd974770486ffb2R717)

pyflakes/checker.py
legau: I would leave it in case of future new named type and because it would repeat 
```python
with self._enter_annotation(AnnotationState.NONE):
    self.handleChildren(node)
```
 and
```python
for a in attributes:
      if isinstance(a, ast.Str):
          with self._enter_annotation():
              self.handleNode(a, node)
```

pyflakes/checker.py
asottile: on your branch:

```python
from typing import NamedTuple
from t import A
from t import C

NamedTuple("A", ("a", A["C"]))
```

```
t.py:3:1 't.C' imported but unused
```

pyflakes/checker.py
asottile: I would prefer the reduced complexity, thanks -- over-DRYing and pre-DRYing make code harder to maintain

pyflakes/checker.py
legau: I guess that would be expected, the forward reference only is checked if pyflakes knows the subscript caller is a member of a typing module. The same error is found with this :
```python
from typing import List
from t import C
A = List
b: A["C"]
```

asottile: that doesn't produce an error -- pyflakes uses the knowledge of a "typing context" to process strings

legau: You're right. However it seems this error already exists in pyflakes current version. Should I improve this PR to manage this ?

asottile: yes, you know which parts of the `NamedTuple` / `TypedDict` expression are typing contexts so you should process them with typing contexts instead of hacky `isinstance` of `str` -- that was my entire point for mentioning this

legau: My new push solves this by keeping a record of annotated nodes which are then always analysed inside a typing context

pyflakes/checker.py
```diff
@@ -1699,7 +1704,33 @@ def CALL(self, node):
...
                         self.handleNode(keyword.value, node)
 
-        self.handleChildren(node)
+        if _is_typing(node.func, "TypedDict", self.scopeStack):
+            # TypedDict("a", {"a": int})
+            if len(node.args) > 1 and isinstance(node.args[1], ast.Dict):
+                self._annotated_nodes += node.args[1].values
+
+            # TypedDict("a", a=int)
+            self._annotated_nodes += [k.value for k in node.keywords]
```
asottile, line 1713: 🤦 just use `self._enter_annotation` inline right here, don't introduce state that's unneeded -- look at how `TypeVar` right above this works

pyflakes/checker.py
```diff
@@ -1699,7 +1704,33 @@ def CALL(self, node):
...
+
+            with self._enter_annotation(AnnotationState.NONE):
+                self.handleChildren(node)
+
+        elif _is_typing(node.func, "NamedTuple", self.scopeStack):
+            # NamedTuple("a", [("a", int)])
+            if len(node.args) > 1 and isinstance(
+                node.args[1], (ast.Tuple, ast.List)
+            ) and all(isinstance(x, (ast.Tuple, ast.List)) and
+                      len(x.elts) == 2 for x in node.args[1].elts):
```
asottile, line 1723: this indentation is all over the place and is unreadable.  please write it like:

```
if (
        len(node.args) >1 and
        isinstance(node.args[1], (ast.Tuple, ast.List)) and
        all(...)
):
    # code here
```

pyflakes/checker.py
```diff
@@ -1699,7 +1704,33 @@ def CALL(self, node):
                     with self._enter_annotation():
                         self.handleNode(keyword.value, node)
 
-        self.handleChildren(node)
+        if _is_typing(node.func, "TypedDict", self.scopeStack):
```
asottile, line 1707: this should be `elif`

pyflakes/checker.py
legau: I can't only use `self._enter_annotation`, I need some info that some nodes are potential types while entering with `AnnotationState.NONE` to prevent flagging all other STR as forward references.
Your facepalm is really uncalled for

asottile: you have that information!  it's the same as the things you're assinging to `self._annotated_nodes`

legau: I know I have that information, but I need to handle nested children nodes differently by their "annotated" property or not. I don't think removing these nodes from the tree would be good practice. How would you make it so they're analyzed just once ?

asottile: follow how `TypeVar` does it?

legau: Following TypeVar implementation doesn't solve this case
```python
from typing import NamedTuple
from t import A
from t import C

NamedTuple("A", [("a", A["C"])])
```

legau: TypeVar in this case does not support it either
```python
from typing import TypeVar
from t import A
from t import C

TypeVar("A", bound=A["C"])
```
```
t.py:3:1 't.C' imported but unused
```
Even though for TypeVar this is an incorrect syntax anyway so it doesn't matter

asottile: I see, looks like TypeVar has a slight bug:

```diff
diff --git a/pyflakes/checker.py b/pyflakes/checker.py
index 57603c8..1d18a1e 100644
--- a/pyflakes/checker.py
+++ b/pyflakes/checker.py
@@ -1689,13 +1689,12 @@ class Checker(object):
         elif _is_typing(node.func, 'TypeVar', self.scopeStack):
             # TypeVar("T", "int", "str")
             for arg in node.args[1:]:
-                if isinstance(arg, ast.Str):
-                    with self._enter_annotation():
-                        self.handleNode(arg, node)
+                with self._enter_annotation():
+                    self.handleNode(arg, node)
 
             # TypeVar("T", bound="str")
             for keyword in node.keywords:
-                if keyword.arg == 'bound' and isinstance(keyword.value, ast.Str):
+                if keyword.arg == 'bound':
                     with self._enter_annotation():
                         self.handleNode(keyword.value, node)
 
```

asottile: #610 for that (if you want to review it)

legau: You can't do that because now you're allowing some nodes to be analyzed 2 times. With your implementation,
```python
from typing import TypeVar, List

TypeVar("A", bound=List["A"])
```

```
t.py:3:25 undefined name 'A'
t.py:3:25 undefined name 'A'
```

asottile: that's existing it seems -- not hard to pick apart the `Call` though and visit each member individually

legau: That's what I wanted to do at first, what would be the best way to do that without almost recreating a `self.handleNode` ?

legau:
I changed it so that each subelement is visited once with correct AnnontationState. I also wrapped in `TypeVar` which has a similar behavior

pyflakes/checker.py
```diff
@@ -1687,19 +1691,68 @@ def CALL(self, node):
...
 
         elif _is_typing(node.func, 'TypeVar', self.scopeStack):
+
             # TypeVar("T", "int", "str")
-            for arg in node.args[1:]:
-                if isinstance(arg, ast.Str):
-                    with self._enter_annotation():
-                        self.handleNode(arg, node)
+            omit += ["args"]
+            annotated += [arg for arg in node.args[1:] if isinstance(arg, ast.Str)]
```
asottile, line 1697: even if they're not strings we need to treat them as annotations -- for instance you might have `List["T"]`

pyflakes/checker.py
```diff
@@ -1678,6 +1678,10 @@ def CALL(self, node):
         ):
             self._handle_string_dot_format(node)
 
+        omit = []
+        annotated = []
+        not_annotated = []
+
         if (
             _is_typing(node.func, 'cast', self.scopeStack) and
```
asottile, line 1686: should we handle `cast` as well? I suspect that also has the same issue (since the first argument is a type)

asottile (commented):
sorry I was so slow to review, I've been really busy lately

pyflakes/checker.py
legau: Cast isn't a named type so you can't use it within a subscript like `List[cast(A, 1)]`. The first argument will be either considered annotated if it's a string or it is be a type that

asottile: here's a test case you can use which demonstrates my point:

```python
from typing import cast
from helpers import SomeGeneric

import queue

cast(SomeGeneric["queue.Queue[int]"], [])
```

legau: Ok I see, just have to make it for everything and not just strings I guess.

legau:
It should be OK with your remarks, do you have any other cases you can think of ?

asottile (approved):

PR #610, opened 2021-01-15
treat TypeVar bound / args fully as types

asottile:


legau:
You can't do that because now you're allowing some nodes to be analyzed 2 times. With your implementation,
```python
from typing import TypeVar, List

TypeVar("A", bound=List["A"])
```

```
t.py:3:25 undefined name 'A'
t.py:3:25 undefined name 'A'
```

asottile:
> 
> 
> You can't do that because now you're allowing some nodes to be analyzed 2 times. With your implementation,
> 
> ```python
> from typing import TypeVar, List
> 
> TypeVar("A", bound=List["A"])
> ```
> 
> ```
> t.py:3:25 undefined name 'A'
> t.py:3:25 undefined name 'A'
> ```

that's existing, not a regression

legau:
What do you mean by existing ? When I try it with current pyflakes I only get one of the errors

legau:
I guess it would for a TypeVar inside an annotated context. All in all it raises the exact same issues as #575

PR #623, opened 2021-03-23
allow turning on run tracing from the command line

terencehonles:
This along with some additional tracing output was used to help with #622.

In the additional changes I added the tracing to also print out the stack information when popping the stack. Eventually I added some more context to which stack was being printed and while I shimmed the information in via `def popStack(self, node=None)` to optionally provide node information for name and file/line context. I was wondering if it would be OK if I were to add a node to each stack created instead of trying to "merge" them via the checker.

For the stack information I printed out if the name was used and if it was available for runtime usage. An example of this would be the following:

```python

if TYPE_CHECKING:
    from module import Type

    class A:
        def a(x) -> Type:
            pass
else:
    class A:
        def a(x):
            pass

def b(x) -> Type:
    from module import Type

    assert isinstance(x, Type)
    return x

class C:
    def c(x) -> Type:
        from module import Type

        assert isinstance(x, Type)
        return x
```

```
ImportFrom
end ImportFrom
If
  Name
  end Name
  ImportFrom
  end ImportFrom
  ClassDef
    FunctionDef
      Name
      end Name
    end FunctionDef
    ClassScope(A):<stdin>:6:5
      a:
        used: False
        runtime: False
  end ClassDef
  ClassDef
    FunctionDef
    end FunctionDef
    ClassScope(A):<stdin>:10:5
      a:
        used: False
        runtime: True
  end ClassDef
end If
FunctionDef
  Name
  end Name
end FunctionDef
ClassDef
  FunctionDef
    Name
    end Name
  end FunctionDef
  ClassScope(C):<stdin>:20:1
    c:
      used: False
      runtime: True
end ClassDef
arguments
  arg
  end arg
end arguments
Pass
end Pass
FunctionScope(a):<stdin>:7:9
  x:
    used: False
    runtime: True
arguments
  arg
  end arg
end arguments
Pass
end Pass
FunctionScope(a):<stdin>:11:9
  x:
    used: False
    runtime: True
arguments
  arg
  end arg
end arguments
ImportFrom
end ImportFrom
Assert
  Call
    Name
    end Name
    Name
    end Name
    Name
    end Name
  end Call
end Assert
Return
  Name
  end Name
end Return
FunctionScope(b):<stdin>:14:1
  Type:
    used: True
    runtime: True
  x:
    used: True
    runtime: True
arguments
  arg
  end arg
end arguments
ImportFrom
end ImportFrom
Assert
  Call
    Name
    end Name
    Name
    end Name
    Name
    end Name
  end Call
end Assert
Return
  Name
  end Name
end Return
FunctionScope(c):<stdin>:21:5
  Type:
    used: True
    runtime: True
  x:
    used: True
    runtime: True
ModuleScope
  A:
    used: False
    runtime: True
  C:
    used: False
    runtime: True
  TYPE_CHECKING:
    used: True
    runtime: True
  Type:
    used: True
    runtime: False
  b:
    used: False
    runtime: True
  isinstance:
    used: True
    runtime: True
```

### DIFF
--- a/pyflakes/api.py
+++ b/pyflakes/api.py
@@ -17,7 +17,7 @@ __all__ = ['check', 'checkPath', 'checkRecursive', 'iterSourceCode', 'main']
 PYTHON_SHEBANG_REGEX = re.compile(br'^#!.*\bpython([23](\.\d+)?|w)?[dmu]?\s')
 
 
-def check(codeString, filename, reporter=None):
+def check(codeString, filename, reporter=None, trace=False):
     """
     Check the Python source given by C{codeString} for flakes.
 
@@ -71,14 +71,15 @@ def check(codeString, filename, reporter=None):
         return 1
     # Okay, it's syntactically valid.  Now check it.
     file_tokens = checker.make_tokens(codeString)
-    w = checker.Checker(tree, file_tokens=file_tokens, filename=filename)
+    w = checker.Checker(
+        tree, file_tokens=file_tokens, filename=filename, trace=trace)
     w.messages.sort(key=lambda m: m.lineno)
     for warning in w.messages:
         reporter.flake(warning)
     return len(w.messages)
 
 
-def checkPath(filename, reporter=None):
+def checkPath(filename, reporter=None, trace=False):
     """
     Check the given path, printing out any warnings detected.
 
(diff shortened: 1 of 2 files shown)


sigmavirus24 (commented):
This project doesn't use flags purposefully. Further `astpretty` does what you're asking very well already

asottile:
agreed, pyflakes doesn't have any flags currently and there's better tools for visualizing the ast

terencehonles:
> agreed, pyflakes doesn't have any flags currently and there's better tools for visualizing the ast

But this is not only about tracing the ast, this is about inspecting pyflake's internal state. Is there a better tool for this currently?

asottile:
> Is there a better tool for this currently?

A debugger maybe?

terencehonles:
Ok, I guess I'm glad I didn't try working on making the tracing more useful.

PR #624, opened 2021-03-23
allow passing tox positional arguments to unittest

terencehonles:
This allows `tox -e py27 -- -k <test name>`.


asottile (approved):
thanks!

PR #625, opened 2021-03-23
remove old and unused "tracing" code

terencehonles:
Removing the tracing lines based on the discussion in #623


asottile (approved):
I'd like to get one more regression-bugfix release of pyflakes out (for 2.3.x), but then I think this is good to be merged 👍

PR #630, opened 2021-04-30
add support for match statement

asottile:


pyflakes/checker.py
```diff
@@ -2383,3 +2383,10 @@ def COMPARE(self, node):
             left = right
 
         self.handleChildren(node)
+
+    MATCH = MATCH_CASE = handleChildren
+    MATCHCLASS = MATCHMAPPING = MATCHSEQUENCE = MATCHVALUE = handleChildren
+
+    def MATCHAS(self, node):
+        self.handleNodeStore(node)
+        self.handleChildren(node)
```
asottile, line 2392: ideally I wouldn't need this -- opened https://bugs.python.org/issue43994

asottile:
note: I did not enable CI for 3.10 yet -- waiting on beta freeze so I don't have to change it multiple times (there's a few exception messages which changed with the versions)

pyflakes/checker.py
cdce8p: I noticed that as well for pylint / astroid. Would have been nice to have, but I guess we now have to wait till 3.11 for it 🤷🏻‍♂️

pyflakes/test/test_match.py
```diff
@@ -0,0 +1,76 @@
...
+                    case {'foo': y}:
+                        print(f'matched {y}')
+        ''')
+
+    def test_match_pattern_matched_class(self):
+        self.flakes('''
+            from a import B
+
+            match 1:
+                case B(x=1) as y:
```
cdce8p, line 36: It should also be possible to bind the value of `x` to a new variable. The same for any positional pattern. Something like this:
```py
match 1:
    case B(0, m, x=o) as y:
        ....
```

cdce8p (commented):
LGTM! Aside from the one case I mentioned below, you should have covered all where variables can be bound.

pyflakes/checker.py
```diff
@@ -2378,3 +2381,18 @@ def COMPARE(self, node):
...
         self.handleChildren(node)
+
+    MATCH = MATCH_CASE = MATCHCLASS = MATCHOR = MATCHSEQUENCE = handleChildren
+    MATCHSINGLETON = MATCHVALUE = handleChildren
+
+    def MATCHAS(self, node):
+        self.handleNodeStore(node)
+        self.handleChildren(node)
+
+    def MATCHMAPPING(self, node):
```
sigmavirus24, line 2392: Minor nit, could we do

```py
MATCHMAPPING = MATCHSTAR = MATCHAS
```

sigmavirus24 (approved):
Non-blocking nit left in-line

PR #632, opened 2021-05-14
Detect typing module attributes with 'import typing as <name>'

### LINKED ISSUE #561: `typing.Literal` is not detected when renamed at import time
I just encountered a false-positive regression reported by PyFlakes 2.2.0, but not 2.1.1: If the *Literal* class from `typing` or `typing_extensions` is renamed at import time, PyFlakes looses track of it and treats any string argument as forward-reference that then cannot be resolved – causing it to emit a *F821 undefined name 'none'* error through flake8.

Minimum example:

```py
import typing as ty
from typing_extensions import Literal as ty_Literal
		
@ty.overload
def request(
		decoder: ty_Literal["none"] = "none",
) -> None:
	...
```

angus-lherrou:
Fixes #561

See https://github.com/pallets/click/pull/1896#discussion_r632645323 for use case.

### DIFF
--- a/pyflakes/checker.py
+++ b/pyflakes/checker.py
@@ -717,6 +717,16 @@ def _is_typing_helper(node, is_name_match_fn, scope_stack):
 
         return False
 
+    def _module_scope_is_typing(name):
+        for scope in reversed(scope_stack):
+            if name in scope:
+                return (
+                    isinstance(scope[name], Importation) and
+                    scope[name].fullName in TYPING_MODULES
+                )
+
+        return False
+
     return (
         (
             isinstance(node, ast.Name) and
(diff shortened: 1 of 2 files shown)


asottile (approved):

pyflakes/test/test_type_annotations.py
```diff
@@ -121,6 +121,23 @@ def f(self, x):  # type: (str) -> str
             def f(self, x): return x
         """)
 
+    def test_aliased_import(self):
```
asottile, line 124: looks like you'll need to guard this test to be python 3 only because it uses `async def`

angus-lherrou: @asottile thanks, I didn't even notice that. fixed

pyflakes/checker.py
```diff
@@ -717,14 +717,24 @@ def _bare_name_is_attr(name):
 
         return False
 
+    def _module_scope_is_typing(name):
```
davidism, line 720: Would this be good to expose as a function that plugins can use. Then tools like `flake8-typing-imports` could use the same solution.

asottile: no, pyflakes does not expose its stack handling nor its non-standard ast parser and it's very specialized for pyflakes

Issue #387, opened 2018-11-04
Treat global dunder names as always used


anntzer:
When using a package versioning helper such as setuptools_scm, it is quite natural to write, in a package's toplevel `__init__.py`, something like
```
from ._version import version as __version__
```
(assuming one uses setuptools_scm's `write_to` option rather than pkg_resources, as the latter can be quite slow).

Currently, pyflakes complains that `__version__` is an unused name.  Obviously, one can placate pyflakes by writing (over two lines) `from . import _version; __version__ = _version.version`, but that seems a bit silly: I would instead suggest that pyflakes consider all global-level dunders as automatically used (in the sense "they have a special or metadata role anyways"), even if they appear not so.

asmeurer:
pyflakes currently requires using `__all__` in `__init__.py` to avoid unused name errors (I'm not sure, but I think it's a good idea for even `__version__` to go in `__all__`).

anntzer:
Actually, I think putting `__version__` in `__all__` is a pretty bad idea: someone who does `from foo import *` (say, in the `bar`) module will now also set `bar.__version__` (as it's a string, it doesn't keep track of its origin module, so third party tools will use that as the `bar` module version.
(Yes, `import *` is a bad idea but we're talking about `__all__` anyways...)

PR #645, opened 2021-07-04
Globally imported dunders are implicitly used.

anntzer:
... to support e.g. importing `__version__` from an autogenerated
submodule.

Closes https://github.com/PyCQA/pyflakes/issues/387.

I chose to skip all dunders (per https://www.python.org/dev/peps/pep-0008/#module-level-dunder-names), but I can also just special-case `__version__`, whatever you prefer.

### DIFF
--- a/pyflakes/checker.py
+++ b/pyflakes/checker.py
@@ -693,6 +693,10 @@ def getNodeName(node):
         return node.name
 
 
+def _is_dunder(name):
+    return name.startswith("__") and name.endswith("__") and len(name) > 4
+
+
 TYPING_MODULES = frozenset(('typing', 'typing_extensions'))
 
 
@@ -1026,7 +1030,9 @@ class Checker(object):
             # Look for imported names that aren't used.
             for value in scope.values():
                 if isinstance(value, Importation):
-                    used = value.used or value.name in all_names
+                    used = (value.used or value.name in all_names
+                            or isinstance(scope, ModuleScope)
+                            and _is_dunder(value.name))
                     if not used:
                         messg = messages.UnusedImport
                         self.report(messg, value.source, str(value))
--- a/pyflakes/test/test_imports.py
+++ b/pyflakes/test/test_imports.py
@@ -154,6 +154,12 @@ class Test(TestCase):
         self.flakes('from .. import fu; assert fu')
         self.flakes('from ..bar import fu as baz; assert baz')
 
+    def test_implicitlyUsedDunderImport(self):
+        self.flakes('from foo import __author__')
+        self.flakes('from foo import version as __version__')
+        self.flakes('from .foo import __author__')
+        self.flakes('from .foo import version as __version__')
+
     def test_redefinedWhileUnused(self):
         self.flakes('import fu; fu = 3', m.RedefinedWhileUnused)
         self.flakes('import fu; fu, bar = 3', m.RedefinedWhileUnused)


asottile:
I'm -1 -- add these to `__all__` like any other variable and they'll be marked as used

anntzer:
As discussed in https://github.com/PyCQA/pyflakes/issues/387#issuecomment-435857561 adding them to `__all__` is not a good idea (you almost certainly don't want them imported if doing a star-import).

asottile:
`import *` is a bad idea anyway

anntzer:
I agree, but that's what `__all__` is fundamentally about...

asottile:
not entirely, it's also for documenting the "public" names of a namespace (tab completion, doc generation, etc.)

anntzer:
Fair, but note that https://www.python.org/dev/peps/pep-0008/#module-level-dunder-names specifically does not consider `__version__` to be worth adding to `__all__`.

asottile:
yeah it also imports `barry_as_FLUFL` -- I don't think it's a serious example

anntzer:
It's not a very serious example, but it's also the only thing (AFAICT) in the reference docs that addresses this point...

sigmavirus24:
I'm also a fairly strong -1 on this. `import *` has been discouraged for the greater part of a decade if not more.

anntzer:
To be clear, this is not really about `import *`, but about how to reuse `__version__` defined in a helper module.

sigmavirus24:
Yes, I understand that. Your argument against what seems to be a common practice is potential use of an uncommon one that's strongly recommended against

anntzer:
Well, another argument given above is that PEP8 explicitly does not add `"__version__"` to `__all__`.  Yes, the example is clearly a bit tongue-in-cheek, but, given its clean history (it was added in a single chunk in https://github.com/python/peps/commit/070a23b4a0f50687614afc87302cfe9ead2fcab0), it seems clear that its author did not consider `"__version__"` as being part of `__all__` and that it's not just an omission.

sigmavirus24:
>  but, given its clean history

This is not a convincing argument. The PEP was a point-in-time capture of how to write code for the standard library which `pycodestyle` attempts to enforce (despite the sometimes split-brain nature of the document), not `pyflakes`. This project looks for high confidence lints that it can report.

> I am more than willing to restrict the special-casing to `__version__` rather than all dunders (I don't really care either way)

There are many projects that use `__about__` or `__metadata__` as a module that they import into `__init__` to then re-export things. Of those projects I believe many have far more "dunder" variables than just `__version__`. Special-casing `__version__` opens the door for PRs to special-case more things. Ignoring everything, opens the doors to bug reports because they stopped using a dunder variable and we didn't report it as unused.

There seems to be little possibility for anything other than an absolute increase in maintainer-burden by accepting some variant of this pull request.

anntzer:
> This project looks for high confidence lints that it can report.

My second point in my previous message is that this is with high confidence an invalid lint, and that it artificially forces a perfectly valid single-statement construct to be split into two statements for the sole purpose of satisfying pyflakes.  (On top of that, the two statements will typically have to be far away from one another as one will be in the imports block whereas the other defines a global variable.)

> an absolute increase in maintainer-burden

I would argue that the increase is not large (and is counterbalanced by a real usability improvement), although I am obviously not a maintainer, so my judgment hardly matters here.

PR #649, opened 2021-07-17
Add end coordinates for column and line number

### LINKED ISSUE #454: End column/lineno
It would be useful to have `end_col` and `end_lineno` attributes on Message objects. The way it would work is that (col, lineno) to (end_col, end_lineno) would represent the range of characters for the error. For instance, for undefined names, the col, lineno already points to the first character, and the end versions would point to the last character in the word. 

I haven't looked yet how easy this is to implement. I wanted to post it here to see if anyone had any thoughts. I might make a pull request for it later (time permitting).

krassowski:
Closes #454.

- [x] Added offset when handling positions in docstrings (same as for the start coordinates)
- [x] Added test for the offset handling, conditional on Python >= 3.8\
- [x] Checked that tests pass locally and no flake8 warnings are introduced


pyflakes/messages.py
```diff
@@ -11,6 +11,8 @@ def __init__(self, filename, loc):
         self.filename = filename
         self.lineno = loc.lineno
         self.col = getattr(loc, 'col_offset', 0)
+        self.end_col = getattr(loc, 'end_col_offset', None)
+        self.end_lineno = getattr(loc, 'end_lineno', None)
```
asmeurer, line 15: As I stated in the issue, I think it would much better to make the default to be equal to the start. That way the type is always an integer, and start = end will typically just do the right thing for most code that would deal with these things.

krassowski: I disagree. In the use-case of editors when we want to highlight a range with wavy underline, providing start = end will make the underline invisible. start = end might be handled as a special case e.g. when a character is missing (and then denoted with another type of marker like an arrow pointing to that location in the editor). Coercing all missing data to that representation prevents the editors from distinguishing when they need to show the arrow marker and when they need to infer the underline range themselves (e.g default it to the closest token or the entire line).

asottile:
I don't want to land this as is, when 3.8 is our minimum targetted version we can revisit

pyflakes/messages.py
asmeurer: If editors want to show a different marker, can't they just check if start == end? 

We could also consider this an off by one type issue where we could make the default to be to select a single character.

jayvdb:
@krassowski 3.8 is our minimum targetted version now.  Could you rebase this PR?

PR #651, opened 2021-08-06
Correct README.rst

dlintin:


README.rst
```diff
@@ -6,7 +6,7 @@ A simple program which checks Python source files for errors.
 
 Pyflakes analyzes programs and detects various errors.  It works by
 parsing the source file, not importing it, so it is safe to use on
-modules with side effects.  It's also much faster.
+modules without any side effects.  It's also much faster.
```
krader1961, line 9: What side-effects have you observed that warrant this change?

asottile:
the text is correct as written

README.rst
dlintin: no side effects, that's why should be changed to "without any".

krader1961: If English is your native language I hope your English teachers are performing a Picard double face-palm. If your native language is not English you should probably avoid making changes to the grammar of English text.

sigmavirus24:
@dlintin I'm sorry for @krader1961's behaviour. It's inappropriate. 

I will say that I think this sentence could be written in a clearer way, even if this is written correctly as it is written today. @dlintin to help, we're saying that "If you have a module which (when imported) has side-effects those won't be taken into account or triggered because pyflakes parses the Abstract Syntax Tree." Does that make sense? If you're up for it, would you be up for sending a new PR to clarify that sentiment for you? It took me a couple reads of this to parse it correctly too just now because I'm particularly exhausted.

PR #655, opened 2021-10-02
Add project URLs to setup.py metadata for PyPI

remram44:
This will make it easier to find e.g. the changelog or mailing-list by adding links to the left bar on [the pypi.org page](https://pypi.org/project/pyflakes/):

PyPI screenshot

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -43,6 +43,12 @@ setup(
     author="A lot of people",
     author_email="code-quality@python.org",
     url="https://github.com/PyCQA/pyflakes",
+    project_urls={
+        "Source": "https://github.com/PyCQA/pyflakes",
+        "Changelog": "https://github.com/PyCQA/pyflakes/blob/master/NEWS.rst",
+        "Bug Tracker": "https://github.com/PyCQA/pyflakes/issues",
+        "Mailing List": "https://mail.python.org/mailman/listinfo/code-quality",
+    },
     packages=["pyflakes", "pyflakes.scripts", "pyflakes.test"],
     python_requires='>=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*',
     classifiers=[


setup.py
```diff
@@ -43,6 +43,12 @@ def get_long_description():
     author="A lot of people",
     author_email="code-quality@python.org",
     url="https://github.com/PyCQA/pyflakes",
+    project_urls={
+        "Source": "https://github.com/PyCQA/pyflakes",
```
asottile, line 47: this duplicates the `url` that's already there -- I don't think this adds any value

setup.py
```diff
@@ -43,6 +43,12 @@ def get_long_description():
...
+        "Changelog": "https://github.com/PyCQA/pyflakes/blob/master/NEWS.rst",
```
asottile, line 48: this is already linked from the readme which displays on pypi.org

setup.py
```diff
@@ -43,6 +43,12 @@ def get_long_description():
...
+        "Bug Tracker": "https://github.com/PyCQA/pyflakes/issues",
```
asottile, line 49: this is also already linked because we used github as the `url` so we don't need this either

setup.py
```diff
@@ -43,6 +43,12 @@ def get_long_description():
...
+        "Mailing List": "https://mail.python.org/mailman/listinfo/code-quality",
```
asottile, line 50: this is also already linked from the readme

asottile (commented):
I actually think more links here distracts from where we want people to click which is through to github

remram44:
I thought that you linked from the README **because** you wanted people to be able to get there easily from the PyPI page. Right now the first link you find when scanning the README is a link to the PyPI page, you have to scroll two pages down to find the changelog and mailing-list. Apologies if I was wrong.

In fact, I got here by clicking on the "GitHub statistics" links that show up on the left, and then browsing back from the "stargazer" page back to the repo index. I suppose I could have checked whether the "Homepage" link was GitHub, but it often isn't, so that's what I usually do when there's no "source" link.

screenshot of GitHub statistics section

If you think this makes it too crowded, no worries, I just thought it would be helpful. I don't think you can prevent the whole "GitHub statistics" section from showing up though, as long as your "url" field points to GitHub.

Issue #461, opened 2019-08-08
Several syntax errors not caught


asmeurer:
I just realized that I never opened an issue about this. After https://github.com/PyCQA/pyflakes/pull/20, several compile-time syntax errors are warned about, but not all of them. The reason is that when I was going through the CPython source to find the compile-time errors, some of them seemed too esoteric or complicated, so I didn't implement them. A noncomprehensive list of example:

- Invalid use of global and nonlocal variables https://github.com/PyCQA/pyflakes/issues/249
- Some star expression things in Python 3.4. I don't know if pyflakes still supports 3.4, or if any of these are left in 3.5.
- Anything relating to async/await

These probably should be split out into separate issues, but it would also be a good idea to do an audit of compile.c (or just use [my original idea](https://github.com/PyCQA/pyflakes/pull/8), which will catch every compile-time syntax error automatically).

PR #666, opened 2021-12-24
Global variables must be assigned to be used

### LINKED ISSUE #590: global keyword removes undefined variable warning
```py
def test():
    return x
```

gives `test.py:2:12 undefined name 'x'` as expected, but 

```py
def test():
    global x
    return x
```

gives no warnings.

dannysepler:
Closes #590

I'm pretty new to `pyflakes` so just let me know if I messed something up


pyflakes/checker.py
```diff
@@ -322,10 +322,11 @@ class Binding(object):
                 the node that this binding was last used.
     """
 
-    def __init__(self, name, source):
+    def __init__(self, name, source, isUnassigned=False):
         self.name = name
         self.source = source
         self.used = False
+        self.isUnassigned = isUnassigned
```
asottile, line 329: new code should use `snake_case` -- we're slowly migrating over the legacy code which uses the old casing

pyflakes/checker.py
```diff
@@ -322,10 +322,11 @@ class Binding(object):
                 the node that this binding was last used.
     """
 
-    def __init__(self, name, source):
+    def __init__(self, name, source, isUnassigned=False):
```
asottile, line 325: this is a bit of a double-negative and a [boolean trap](https://youtu.be/CnRkXO_a5mI) -- it would probably be best to change this to `assigned=True` and make it keyword-only

pyflakes/checker.py
```diff
@@ -1129,6 +1130,10 @@ def addBinding(self, node, value):
                 break
         existing = scope.get(value.name)
 
+        in_global_scope = value.name in self.scopeStack[-1]
+        if existing and in_global_scope and existing.isUnassigned:
+            existing.isUnassigned = False
```
asottile, line 1135: I don't think this is correct -- for instance a closure-scoped variable can shadow a global variable (and is a separate scope)

pyflakes/checker.py
```diff
@@ -1207,6 +1212,12 @@ def handleNodeLoad(self, node):
                     continue
 
             binding = scope.get(name, None)
+            try:
+                if binding.isUnassigned:
+                    self.report(messages.UndefinedName, node, name)
+            except AttributeError:
+                pass
```
asottile, line 1219: we shouldn't be blindly catching AttributeError -- this is a programming error

pyflakes/test/test_undefined_names.py
```diff
@@ -335,6 +335,21 @@ def c(): bar
         def b(): global bar; bar = 1
         ''')
 
+    def test_globalIsUndefinedIfNeverGivenValue(self):
+        """
+        If a "global" is never given a value, it is undefined
+        """
+        self.flakes('''
```
asottile, line 342: I don't think this is necessarily true, a function later in the module could modify that variable and give it a value

pyflakes/test/test_undefined_names.py
```diff
@@ -366,15 +381,16 @@ def f2():
...
+        """
         self.flakes('''
         def f1():
-            m
+            global m
 
         def f2():
-            global m
+            m
         ''', m.UndefinedName)
```
asottile, line 394: this test shouldn't have been altered -- it was demonstrating a behaviour that *should* happen

pyflakes/test/test_undefined_names.py
```diff
@@ -499,7 +515,7 @@ def fun2():
                     a
                     a = 2
                     return a
-        ''', m.UndefinedLocal)
+        ''', m.UndefinedLocal, m.UndefinedName)
```
asottile, line 518: I don't think this new error here is correct -- it seems to be correct on the main branch

pyflakes/checker.py
dannysepler: sure! should i wait for https://github.com/PyCQA/pyflakes/pull/660 to be merged before introducing a keyword-only arg?

asottile: yeah I just merged it

pyflakes/test/test_undefined_names.py
dannysepler: at the moment, this is the one part of the diff i'm still iffy about. you're right, this should eventually just be un-skipped. `pyflakes` at the moment only errors when:

```python
def f1():
    global m

def f2():
    m
```

and it does not error if the global is defined in `f2`. not sure if this is some bigger thing with pyflakes or if it's a simple fix i can do in here!

pyflakes/checker.py
dannysepler: sweet, added a test case for that

pyflakes/checker.py
```diff
@@ -1155,6 +1162,10 @@ def handleNodeLoad(self, node):
                     continue
 
             binding = scope.get(name, None)
+
+            if getattr(binding, 'assigned', None) is False:
```
asottile, line 1166: ah -- my point on this was more that this shouldn't be something we're guessing at -- it should always have this attribute or never have this attribute (it should be set unconditionally on construction)

pyflakes/checker.py
```diff
@@ -1073,6 +1074,12 @@ def addBinding(self, node, value):
                 break
         existing = scope.get(value.name)
 
+        global_scope = self.scopeStack[-1]
+        if (existing and global_scope.get(value.name) == existing and
+                not existing.assigned):
+            # make sure the variable is in the global scope before setting as assigned
+            existing.assigned = True
```
asottile, line 1081: pyflakes running on itself is showing this is probably not quite  right -- or at least it's not happy about `nonlocal`

dannysepler:
pyflakes is treating `nonlocal` the same as `globals` at the moment, this is starting to fork the two! it doesn't solve https://github.com/PyCQA/pyflakes/issues/249 or https://github.com/PyCQA/pyflakes/issues/461 yet, but at least it opens us up to solving it later?

PR #667, opened 2021-12-25
Tear down a bunch of py2 logic

dannysepler:
This follows https://github.com/PyCQA/pyflakes/pull/660 -- these were a bit distracting to look at

If this goes through, I can tear down <= py3.5 checks

### DIFF
--- a/pyflakes/checker.py
+++ b/pyflakes/checker.py
@@ -19,7 +19,6 @@ import tokenize
 
 from pyflakes import messages
 
-PY2 = sys.version_info < (3, 0)
 PY35_PLUS = sys.version_info >= (3, 5)    # Python 3.5 and above
 PY36_PLUS = sys.version_info >= (3, 6)    # Python 3.6 and above
 PY38_PLUS = sys.version_info >= (3, 8)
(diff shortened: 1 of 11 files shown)


pyflakes/checker.py
```diff
@@ -2118,24 +2079,14 @@ def LAMBDA(self, node):
...
+            for arg in node.args.posonlyargs:
                 args.append(arg.arg)
                 annotations.append(arg.annotation)
-            defaults = node.args.defaults + node.args.kw_defaults
+        for arg in node.args.args + node.args.kwonlyargs:
+            args.append(arg.arg)
+            annotations.append(arg.annotation)
+        defaults = node.args.defaults + node.args.kw_defaults
 
         # Only for Python3 FunctionDefs
```
dannysepler, line 2091: this comment isn't totally true. i think Lambdas also get caught in this if, but since I didn't totally know what was going on I didn't want to reword it yet

asottile:
I believe another contributor is already working on this -- I'd like to give them the chance first since they've had a difficult time integrating their branch due to conflicts -- plus they used [pyupgrade](https://github.com/asottile/pyupgrade) which gets a lot more things automatically

dannysepler:
sounds good to me! just lemme know if you'd like me to reopen this

PR #668, opened 2021-12-25
Detect unused annotations in functions

### LINKED ISSUE #546: Is an unused variable annotation able to be detected?
Given this sample file:

```python
def foo():
    a = 10
    b: str
    return 1


foo()
```

It correctly gives an error about the variable `a`: `test.py:2:5 local variable 'a' is assigned to but never used`

It would be nice if an error could be given about `b` as well since it is also never used. I'm not sure if this is a bug report or a feature request, but I just wasn't sure if this was possible.

dannysepler:
Closes #546

### DIFF
--- a/pyflakes/checker.py
+++ b/pyflakes/checker.py
@@ -588,7 +588,7 @@ class FunctionScope(Scope):
         self.returnValue = None     # First non-empty return
         self.isGenerator = False    # Detect a generator
 
-    def unusedAssignments(self):
+    def unused_assignments(self):
         """
         Return a generator for the assignments which have not been used.
         """
@@ -600,15 +600,12 @@ class FunctionScope(Scope):
                     isinstance(binding, Assignment)):
                 yield name, binding
 
-    def unusedAnnotations(self):
+    def unused_annotations(self):
         """
         Return a generator for the annotations which have not been used.
         """
         for name, binding in self.items():
-            if (not binding.used and
-                    name not in self.globals and
-                    not self.usesLocals and
-                    isinstance(binding, Annotation)):
+            if not binding.used and isinstance(binding, Annotation):
                 yield name, binding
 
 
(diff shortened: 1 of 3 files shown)


pyflakes/checker.py
```diff
@@ -1208,6 +1219,7 @@ def handleNodeLoad(self, node):
 
             binding = scope.get(name, None)
             if isinstance(binding, Annotation) and not self._in_postponed_annotation:
+                scope[name].used = True
```
dannysepler, line 1222: i only really need this to be truthy, but should i define it as a tuple, like in the `try` below?

pyflakes/checker.py
```diff
@@ -600,6 +600,17 @@ def unusedAssignments(self):
...
                 yield name, binding
 
+    def unusedAnnotations(self):
+        """
+        Return a generator for the annotations which have not been used.
+        """
+        for name, binding in self.items():
+            if (not binding.used and
+                    name not in self.globals and
+                    not self.usesLocals and
```
asottile, line 610: I know this is copied from above, but are these still relevant?  does annotating a global do anything? does `locals()` include annotated-only things?

pyflakes/messages.py
```diff
@@ -156,6 +156,25 @@ def __init__(self, filename, loc, names):
...
+    used.
+    """
+    message = 'local variable %r is annotated but never used'
+
+    def __init__(self, filename, loc, names):
+        Message.__init__(self, filename, loc)
+        self.message_args = (names,)
+
+
+class ReturnWithArgsInsideGenerator(Message):
```
asottile, line 171: merge conflict was resolved incorrectly it looks like

pyflakes/test/test_type_annotations.py
```diff
@@ -356,18 +356,24 @@ def test_unused_annotation(self):
         class Cls:
             y: int
         ''')
-        # TODO: this should print a UnusedVariable message
+        # This should print an UnusedAnnotation message
```
asottile, line 359: can just delete the comment now \o/

pyflakes/checker.py
```diff
@@ -600,6 +600,17 @@ def unusedAssignments(self):
                     isinstance(binding, Assignment)):
                 yield name, binding
 
+    def unusedAnnotations(self):
```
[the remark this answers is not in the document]
dannysepler: I felt awkward having `unused_annotations` be snake case while `unusedAssignments` was pascal case, so I made them both snake case

pyflakes/checker.py
dannysepler: Nah I like your point. Seems like it can be simpler, yet still do what we need?

asottile (approved):

PR #672, opened 2022-01-24
Fix path variable name

tushar-deepsource:
`args` I believe is being wrongly shadowed here.


asottile:
sorry we generally don't accept cosmetic changes unless it fixes something -- it doesn't appear to fix anything here (plus `path` is the wrong name, it's a list)

Issue #677, opened 2022-02-06
No duplicate keys detected for variable attributes


frrakn:
Duplicate dict keys will not be detected if they are attributes of a variable.

Popular example is enums. As is, the below code will NOT raise any linter issues.
```
class SomeEnum(Enum):
    FOO: 1

d = {
    SomeEnum.FOO: 1
    SomeEnum.FOO: 2
}
```

asottile:
via https://github.com/PyCQA/pyflakes/pull/678#issuecomment-1030767197

PR #678, opened 2022-02-06
Also detect duplicate dictionary keys if they are variable attributes

frrakn:
See issue #677.


asottile:
it is only meant to detect *literal* maps -- expressions can't necessarily be determined to be identical:

```python
class Next:
    def __init__(self):
        self._x = 0

    def __getattr__(self, _):
        self._x += 1
        return self._x

N = Next()
print({N.x: 0, N.x: 1})
```

Issue #671, opened 2022-01-20
"imported but unused" for `TypeAlias` with string literal type


lkadian:
[Explicit Type Aliases](https://www.python.org/dev/peps/pep-0613/) with type specified as a string literal are incorrectly flagged as an unused import:

**foo.py**
```python
import sys
from os import PathLike

if sys.version_info[:2] >= (3, 10):
    from typing import TypeAlias
else:
    from typing_extensions import TypeAlias

PathLikeStr: TypeAlias = "PathLike[str]"
```

```bash
$ pyflakes --version
2.4.0 Python 3.8.6 on Linux
$ pyflakes foo.py
foo.py:2:1 'os.PathLike' imported but unused
```

Types in string literals are correctly detected in other scenarios:

**ok.py**
```python
from os import PathLike

def f(path: "PathLike[str]") -> "PathLike[str]":
    return path
```

```bash
$ pyflakes ok.py
$
```

asottile:
pyflakes likely needs special handling to understand `TypeAlias` -- it would go here: https://github.com/PyCQA/pyflakes/blob/dbb18432165bdc567e015511b3f3284f0f14cf05/pyflakes/checker.py#L2374

would you be interested in contributing towards that?

lkadian:
Yes I'd be interested in contributing! I'll give it a go

PR #679, opened 2022-02-13
support `TypeAlias` annotations

lkadian:
For assignments with `TypeAlias` as annotation, handle the value as an annotation also.

Avoids incorrectly detecting `Bar` as unused in `x: TypeAlias = 'Bar'`.

resolves https://github.com/PyCQA/pyflakes/issues/671

### DIFF
--- a/pyflakes/checker.py
+++ b/pyflakes/checker.py
@@ -128,7 +128,7 @@ def _is_const_non_singleton(node):  # type: (ast.AST) -> bool
     return _is_constant(node) and not _is_singleton(node)
 
 
-def _is_name_or_attr(node, name):  # type: (ast.Ast, str) -> bool
+def _is_name_or_attr(node, name):  # type: (ast.AST, str) -> bool
     return (
         (isinstance(node, ast.Name) and node.id == name) or
         (isinstance(node, ast.Attribute) and node.attr == name)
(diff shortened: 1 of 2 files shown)


pyflakes/test/test_type_annotations.py
```diff
@@ -300,6 +300,59 @@ def g(t: 'T'): pass
         a: 'a: "A"'
         ''', m.ForwardAnnotationSyntaxError)
 
+    @skipIf(version_info < (3,), 'new in Python 3')
```
asottile, line 303: pyflakes only targets python3.6+

pyflakes/checker.py
```diff
@@ -2381,9 +2388,13 @@ def EXCEPTHANDLER(self, node):
     def ANNASSIGN(self, node):
         self.handleNode(node.target, node)
         self.handleAnnotation(node.annotation, node)
+        # If the assignment has value, handle the *value* now.
         if node.value:
-            # If the assignment has value, handle the *value* now.
-            self.handleNode(node.value, node)
+            # If the annotation is `TypeAlias`, handle the *value* as an annotation.
+            if _is_name_or_const(node.annotation, 'TypeAlias'):
```
asottile, line 2394: this should use `is_typing_helper` which is specifically designed for this

lkadian: thanks, updated

pyflakes/checker.py
```diff
@@ -739,6 +739,10 @@ def _module_scope_is_typing(name):
             isinstance(node.value, ast.Name) and
             _module_scope_is_typing(node.value.id) and
             is_name_match_fn(node.attr)
+        ) or (
+            isinstance(node, ast.Constant) and
```
lkadian, line 743: fyi, mypy understands `x: 'TypeAlias' = Foo` (https://mypy-play.net/?mypy=latest&python=3.10&gist=f7ee6de9eea2f197d6c43039acd97ead)

although I didn't see this in the spec https://www.python.org/dev/peps/pep-0613/ I thought it could be supported also

asottile: let's skip that complexity for now -- it seems very unlikely that someone would do that

asottile (approved):

PR #684, opened 2022-03-25
remove handling of `# type:` comments now that pyflakes is py3+

### LINKED ISSUE #683: Use `ast.parse(type_comments=True)` instead of `file_tokens` on Python 3.8+
Python 3.8 added the ability to include type comments in the abstract syntax tree.

Currently pyflakes uses `tokenize` to find the comments, which is slow. Using type_comments from `ast` when available would be much faster.

asottile:
resolves #683


asottile:
going to merge this for now -- might revert it before the next release but I'd like to see it go out

eli-schwartz:
Just hit this when flake8 updated in CI, it gave me the right kick to finally update this old code... (we use modern annotation style in new code, of course).

In case anyone else has the same issue, https://github.com/ilevkivskyi/com2ann seems useful to auto-migrate your code, so the change should be low-impact.

PR #698, opened 2022-05-24
assignment expression in comprehension should target outer scope

### LINKED ISSUE #633: assignment expression in comprehension should target outer scope
Please consider the following example, which finds the first String in a list that contains the Substring "OH":

```
list = ["STEVE", "JOHN", "YOANN"]
pattern = re.compile(".*%s.*" % "OH")
word = ""

if any((match := pattern.match(item)) for item in list):
    word = match.group(0)
print(word)
```

The code works as intended and outputs "JOHN", but I am getting the following warning from flake8 at 
the line ```word = match.group(0):```

```
F821 -- undefined name 'match'
```

I made a post on [StackOverflow](https://stackoverflow.com/questions/67547135/why-does-flake8-give-the-warning-f821-undefined-name-match-when-using-match), where Anthony Sottile recommended me to post the issue here. I'm using flake8 3.9.2 (mccabe: 0.6.1, pycodestyle: 2.7.0, pyflakes: 2.3.1) CPython 3.9.5 on Windows.

yannsartori:
Closes #633

I wasn't sure where to put the test, so I put it next to the other named expression test

### DIFF
--- a/pyflakes/checker.py
+++ b/pyflakes/checker.py
@@ -540,6 +540,12 @@ class Assignment(Binding):
     """
 
 
+class NamedExprAssignment(Assignment):
+    """
+    Represents binding a name with an assignment expression.
+    """
+
+
 class Annotation(Binding):
     """
     Represents binding a name to a type without an associated value.
@@ -1160,6 +1166,14 @@ class Checker(object):
         # in scope
         if value.name not in self.scope or not isinstance(value, Annotation):
             self.scope[value.name] = value
+            scope = self.scope
+            # As per PEP 572, use scope in which Generator is defined
+            if (
+                isinstance(value, NamedExprAssignment) and
+                isinstance(self.scope, GeneratorScope)
+            ):
+                scope = self.scopeStack[-2]
+            scope[value.name] = value
 
     def _unknown_handler(self, node):
         # this environment variable configures whether to error on unknown
@@ -1303,7 +1317,10 @@ class Checker(object):
         elif PY2 and isinstance(getattr(node, 'ctx', None), ast.Param):
             binding = Argument(name, self.getScopeNode(node))
         else:
-            binding = Assignment(name, node)
+            binding = (
+                Assignment(name, node) if not isinstance(parent_stmt, ast.NamedExpr)
+                else NamedExprAssignment(name, node)
+            )
         self.addBinding(node, binding)
 
     def handleNodeDelete(self, node):
--- a/pyflakes/checker.py
+++ b/pyflakes/checker.py
@@ -1165,7 +1165,6 @@ class Checker(object):
         # don't treat annotations as assignments if there is an existing value
         # in scope
         if value.name not in self.scope or not isinstance(value, Annotation):
-            self.scope[value.name] = value
             scope = self.scope
             # As per PEP 572, use scope in which Generator is defined
             if (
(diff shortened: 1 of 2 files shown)


pyflakes/checker.py
```diff
@@ -1303,7 +1317,10 @@ def handleNodeStore(self, node):
         elif PY2 and isinstance(getattr(node, 'ctx', None), ast.Param):
             binding = Argument(name, self.getScopeNode(node))
         else:
-            binding = Assignment(name, node)
+            binding = (
+                Assignment(name, node) if not isinstance(parent_stmt, ast.NamedExpr)
+                else NamedExprAssignment(name, node)
+            )
```
asottile, line 1323: I would write this as an elif ... else

pyflakes/checker.py
```diff
@@ -1160,6 +1166,14 @@ def addBinding(self, node, value):
         # in scope
         if value.name not in self.scope or not isinstance(value, Annotation):
             self.scope[value.name] = value
+            scope = self.scope
+            # As per PEP 572, use scope in which Generator is defined
+            if (
+                isinstance(value, NamedExprAssignment) and
+                isinstance(self.scope, GeneratorScope)
+            ):
+                scope = self.scopeStack[-2]
```
asottile, line 1175: does this need special handling for `nonlocal` / `global` ?

pyflakes/checker.py
```diff
@@ -1160,6 +1166,14 @@ def addBinding(self, node, value):
         # in scope
         if value.name not in self.scope or not isinstance(value, Annotation):
             self.scope[value.name] = value
+            scope = self.scope
```
asottile, line 1169: if I'm reading this correctly -- it's now assigning twice?

yannsartori: @asottile You were... That assignment shouldn't have been there; I removed it.

pyflakes/checker.py
yannsartori: @asottile I don't think so? I think the caveats pointed out in the PEP about nonlocal/global regard which variables get assigned. But since pyflakes just does static analysis, I don't think it is relevant. I tested it with the following snippet:
```python
global_var = False


def global_keyword_test():
    global global_var
    if any(global_var := f[0] for f in [[True]]):
        print(global_var)


def global_no_keyword_test():
    if any(global_var := f[0] for f in [[True]]):
        print(global_var)


def nonlocal_keyword_test():
    local_var = False

    def helper():
        nonlocal local_var
        if any(local_var := f[0] for f in [[True]]):
            print(local_var)
    helper()
    print(local_var)


def nonlocal_no_keyword_test():
    local_var_2 = False

    def helper():
        if any(local_var_2 := f[0] for f in [[True]]):
            print(local_var_2)
    helper()
    print(local_var_2)


def original_issue():
    def helper():
        if any(local_var_v2 := f[0] for f in [[True]]):
            print(local_var_v2)
    helper()


print("global_keyword_test")
global_keyword_test()
print(global_var)
print("")
global_var = False

print("global_no_keyword_test")
global_no_keyword_test()
print(global_var)
print("")
global_var = False

print("nonlocal_keyword_test")
nonlocal_keyword_test()
print("")

print("nonlocal_no_keyword_test")
nonlocal_no_keyword_test()
print("")
```
And neither `master`, nor this branch, had any issues with any of the functions, except for `master`, which reported the F821 for `original_issue`

The output was as follows:
```
global_keyword_test
True
True

global_no_keyword_test
True
False

nonlocal_keyword_test
True
True

nonlocal_no_keyword_test
True
False
```

asottile (approved):

PR #710, opened 2022-06-23
Allow explicit reexports

harahu:


asottile:
see https://github.com/PyCQA/pyflakes/issues?q=is%3Aissue+reexports+is%3Aclosed

Issue #357, opened 2018-08-12
Some syntax errors are not reported when linting from stdin


stevenkaras:
Minimal reproduction (at least for Python 2.7.12 on Ubuntu 16.04 using pyflakes 2.0.0):

```python
max(1 for i in range(10), key=lambda x: x+1)
```

will not report the syntax error "Generator expression must be parenthesized if not sole argument" if piped into stdin as follows:

```bash
echo 'max(1 for i in range(10), key=lambda x: x+1)' | pyflakes
```

instead failing with a decoding error. However, this is not handled well by many integrations which parse pyflakes output for a lineno and offset.

It looks like there was an assumption in 2009 (from b73a3d12) that if `text is None`, then it's necessarily a decoding issue. I've yet to debug exactly why this specific class of syntax error does not include the text when reading from stdin.

EDIT: just confirmed this affects python3 as well (using Python 3.7.0 on Alpine 3.8 using pyflakes 2.0.0)

myint:
Seems that if I put in any arbitrary valid filename into `ast.parse()`, it properly raises the syntax error. On the other hand, if the file does not exist, it does the current behavior.

https://github.com/PyCQA/pyflakes/blob/0b163640274b15b74ac9a650cf24439b0205fc76/pyflakes/api.py#L41

stevenkaras:
Interesting. To be more accurate, it does not correctly raise the syntax error, but rather reads the text from a file with the given filename if it exists, which need not have anything to do with the actual code you passed into `ast.parse`.

In any case, it seems to me that pyflakes should not assume that if `text is None` that it's an encoding error. Perhaps it would be enlightening to see under what conditions python's builtin `compile()` doesn't include the text field?

For context, I discovered this because SublimeLinter pipes its buffer into pyflakes on stdin, and silently drops all reported issues that don't have a FILENAME:LINE:POSITION format.

Hanaasagi:
When raise `SyntaxError`, it will read the code that have invalid syntax from file. At least, if file is not exist, `text` field will be `None`.

https://github.com/python/cpython/blob/c6b31061997526b31961ec34328408ca421f51fc/Python/ast.c#L681-L685

```C
    loc = PyErr_ProgramTextObject(c->c_filename, LINENO(n));
    if (!loc) {
        Py_INCREF(Py_None);
        loc = Py_None;
    }
```

[...]

PR #716, opened 2022-07-13
fix syntax error reporting from stdin (#357)

stevenkaras:
In b73a3d12, there was an assumption that text is None only if there was an
encoding error with the file. However this was the case for all pythons before
3.9 when reading code from stdin.

This takes care to correctly report as much context as possible, so errors
aren't silently dropped with the unhelpful "problem decoding source" message.

### DIFF
--- a/pyflakes/test/test_api.py
+++ b/pyflakes/test/test_api.py
@@ -15,6 +15,7 @@ from pyflakes.messages import UnusedImport
 from pyflakes.reporter import Reporter
 from pyflakes.api import (
     main,
+    check,
     checkPath,
     checkRecursive,
     iterSourceCode,
@@ -254,6 +255,17 @@ class TestReporter(TestCase):
             ("foo.py:3: a problem\n"
              "bad line of source\n"),
             err.getvalue())
+
+    def test_syntaxErrorNoText(self):
+        """
+        C{syntaxError} doesn't include text or nonsensical offsets if C{text} is C{None}.
+
+        This typically happens when reporting syntax errors from stdin.
+        """
+        err = io.StringIO()
+        reporter = Reporter(None, err)
+        reporter.syntaxError('<stdin>', 'a problem', 0, 0, None)
+        self.assertEqual(("<stdin>:1:1: a problem\n"), err.getvalue())
 
     def test_multiLineSyntaxError(self):
         """
@@ -606,7 +618,8 @@ x = "%s"
 """ % SNOWMAN).encode('utf-8')
         with self.makeTempFile(source) as sourcePath:
             self.assertHasErrors(
-                sourcePath, [f"{sourcePath}: problem decoding source\n"])
+                sourcePath,
+                [f"{sourcePath}:1:1: 'ascii' codec can't decode byte 0xe2 in position 21: ordinal not in range(128)\n"])  # noqa: E501
 
     def test_misencodedFileUTF16(self):
         """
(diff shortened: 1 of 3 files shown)


pyflakes/test/test_api.py
```diff
@@ -606,7 +606,11 @@ def test_misencodedFileUTF8(self):
 """ % SNOWMAN).encode('utf-8')
         with self.makeTempFile(source) as sourcePath:
             self.assertHasErrors(
-                sourcePath, [f"{sourcePath}: problem decoding source\n"])
+                sourcePath,
+                [f"""\
+{sourcePath}:0:0: 'ascii' codec can't decode byte 0xe2 in position 21: ordinal not in \
+range(128)
```
asottile, line 612: rather than escaping this I would just add a `# noqa: E501` comment

pyflakes/test/test_api.py
```diff
@@ -606,7 +606,11 @@ def test_misencodedFileUTF8(self):
 """ % SNOWMAN).encode('utf-8')
         with self.makeTempFile(source) as sourcePath:
             self.assertHasErrors(
-                sourcePath, [f"{sourcePath}: problem decoding source\n"])
+                sourcePath,
+                [f"""\
+{sourcePath}:0:0: 'ascii' codec can't decode byte 0xe2 in position 21: ordinal not in \
```
asottile, line 611: this should probably be `1:1` rather than `0:0` since there is no line number 0

asottile (commented):
could you add a test for this?

stevenkaras:
Added explicit tests of the new behavior and the stdin example as described in the issue.

pyflakes/test/test_api.py
```diff
@@ -255,6 +256,20 @@ def test_syntaxErrorNoOffset(self):
...
+    def test_syntaxErrorNoText(self):
+        """
+        C{syntaxError} doesn't include text or nonsensical offsets if C{text} is C{None}.
+
+        This typically happens when checking input from stdin.
+        """
+        err = io.StringIO()
+        reporter = Reporter(None, err)
+        reporter.syntaxError('<stdin>', 'a problem', 0, 0,
+                             None)
```
asottile, line 268: this should generate an actual syntax error and validate it so we can know when to delete this behaviour (3.9+)

pyflakes/test/test_api.py
```diff
@@ -648,6 +664,43 @@ def test_checkRecursive(self):
...
 
+    def test_stdinReportsErrors(self):
+        """
+        L{check} reports syntax errors from stdin
+        """
+        source = "max(1 for i in range(10), key=lambda x: x+1)\n"
+        err = io.StringIO()
+        count = withStderrTo(err, check, source, "<stdin>")
+        self.assertEqual(count, 1)
+        errlines = err.getvalue().split("\n")[:-1]
```
asottile, line 675: I don't think this actually tests the behaviour -- and we already have a SyntaxError test

pyflakes/test/test_api.py
stevenkaras: This is testing the reporting of a syntax error. Specifically, the behavior of what to do when the offsets are nonsensical (which they are for actual decoding errors in all versions of python including 3.9+). I'll attempt to clarify this in the test comment.

pyflakes/test/test_api.py
stevenkaras: The behavior is triggered when ast.parse throws a SyntaxError from code with a filepath that doesn't exist (stdin).

I'll see if I can trigger this with a different syntax error that has more consistent messaging to avoid the rats nest of expected messages by version. The offset behavior is a bit concerning though. It looks like pypy has special handling that is objectively wrong in certain situations (such as this one).

pyflakes/test/test_api.py
asottile: ah I see -- also I think the `None` fits on the line above within the character limit too if you want to move that as well!  ideally we should set up some auto-formatting

pyflakes/test/test_api.py
asottile: pypy is usually pretty attentive to fixes for syntax error offsets (with a little help from me they inspired the improvements to cpython 3.10's error messages)

stevenkaras: I meant pyflakes handling of pypy's offsets as here: https://github.com/PyCQA/pyflakes/blob/44ef321b0e608c61182ecf5d88f9edeececa5403/pyflakes/api.py#L46-L56

It's probably something that used to be an issue with pypy and was fixed since on the other end. I tested with pypy3.7-7.3.9, but tracking down which version of pypy changed offsets in syntax errors is out of scope for here.

asottile: oh yeah we should totally kill that -- forgot pyflakes had special cases (probably from like 2015-era pypy)

asottile (approved):

PR #723, opened 2022-08-14
Remove universal wheel, python 2 is unsupported

gopackgo90:


asottile:
I use this marker as a "is pure python" -- we use python_requires which is the actual metadata that controls this

gopackgo90:
What do you think about this?:

```
[bdist_wheel]
python-tag = py3
```

I guess I'm not sure if `universal = 1` is some sort of magic "pure python" marker. I'm not familiar with that idea.

asottile:
yeah `universal` means single source pure python

gopackgo90:
https://peps.python.org/pep-0425/#id1
> a user might accept only the *-none-any tags to only download built packages that advertise themselves as being pure Python.

If you were to drop the `universal` flag the file the wheel would be "pyflakes-x.y.z-py3-none-any.whl" instead of "pyflakes-x.y.z-py2.py3-none-any.whl", so the "-none-any" in there still seems to be how PEP 425 defines a pure python wheel.

asottile:
yes however that isn't known statically, and the py2 part is harmless

PR #732, opened 2022-09-16
Remove py2 references from README

dannysepler:
Since https://github.com/PyCQA/pyflakes/pull/707, I don't think pyflakes is runnable from a py2 context anymore

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -23,7 +23,12 @@ It can be installed with::
 
 Useful tips:
 
-* You can also invoke Pyflakes with ``python3.# -m pyflakes .`` (for example, ``python3.10 -m pyflakes .``) if you want to run it for a specific python version.
+* Be sure to install it for a version of Python which is compatible
+  with your codebase: ``python#.# -m pip install pyflakes`` (for example,
+  ``python3.10 -m pip install pyflakes``)
+
+* You can also invoke Pyflakes with ``python#.# -m pyflakes .`` if you want
+  to run it for a specific python version.
 
 * If you require more options and more flexibility, you could give a
   look to Flake8_ too.
--- a/README.rst
+++ b/README.rst
@@ -23,12 +23,7 @@ It can be installed with::
 
 Useful tips:
 
-* Be sure to install it for a version of Python which is compatible
-  with your codebase: for Python 2, ``pip2 install pyflakes`` and for
-  Python3, ``pip3 install pyflakes``.
-
-* You can also invoke Pyflakes with ``python3 -m pyflakes .`` or
-  ``python2 -m pyflakes .`` if you have it installed for both versions.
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -23,12 +23,7 @@ It can be installed with::
 
 Useful tips:
 
-* Be sure to install it for a version of Python which is compatible
-  with your codebase: for Python 2, ``pip2 install pyflakes`` and for
-  Python3, ``pip3 install pyflakes``.
-
-* You can also invoke Pyflakes with ``python3 -m pyflakes .`` or
-  ``python2 -m pyflakes .`` if you have it installed for both versions.
+* You can invoke Pyflakes with ``python -m pyflakes .``
```
asottile, line 26: I think this can be removed entirely -- or adjusted to use something like `python3.10 -m pip install ...` and `python3.10 -m pyflakes ...`

dannysepler: Of the two, I'm maybe team delete since i'd rather not add another thing for maintainers to update in the future

That said, in the `setup.py` it seems like pyflakes always adds some entrypoint, so what if we said:

> You can invoke Pyflakes with `pyflakes .`

asottile: could use `python3.# ...` so it doesn't need updating again.  the original intent of this was to show how to run it per interpreter -- there isn't really much point in showing `pyflakes .` though

dannysepler: `python3.# ...` works for me!

README.rst
```diff
@@ -23,12 +23,7 @@ It can be installed with::
 
 Useful tips:
 
-* Be sure to install it for a version of Python which is compatible
-  with your codebase: for Python 2, ``pip2 install pyflakes`` and for
-  Python3, ``pip3 install pyflakes``.
-
-* You can also invoke Pyflakes with ``python3 -m pyflakes .`` or
-  ``python2 -m pyflakes .`` if you have it installed for both versions.
+* You can invoke Pyflakes with ``python3.# -m pyflakes .``
```
asottile, line 26: ah I meant to keep the original messaging

README.rst
dannysepler: Changed my mind, removing seems simpler. Also, the `flake8` part of the section is implied in `Design Principles` so took out the "Useful tips" section

asottile: this section is important, I want you to adjust it to still make sense but without python 2 -- no reason to throw out the baby with the bathwater

dannysepler: okay gave it another shot, appreciate your patience!

README.rst
```diff
@@ -23,12 +23,7 @@ It can be installed with::
 
 Useful tips:
 
-* Be sure to install it for a version of Python which is compatible
-  with your codebase: for Python 2, ``pip2 install pyflakes`` and for
-  Python3, ``pip3 install pyflakes``.
```
asottile, line 28: this first part is still important but should be replaced with `python#.# -m pip ...`

asottile (approved):

PR #734, opened 2022-10-18
Add __debuggerskip__ as special local

li-dan:
__debuggerskip__ is a special variable used by IPython, similar to __tracebackhide__.

### DIFF
--- a/pyflakes/checker.py
+++ b/pyflakes/checker.py
@@ -567,7 +567,7 @@ class FunctionScope(Scope):
     """
     usesLocals = False
     alwaysUsed = {'__tracebackhide__', '__traceback_info__',
-                  '__traceback_supplement__'}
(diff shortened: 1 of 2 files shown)


asottile:
I believe `PYFLAKES_BUILTINS` or in flake8 `--builtins` already handles this?

li-dan:
I don't think so; `PYFLAKES_BUILTINS` allows suppressing "undefined name" errors, but this is an "assigned to but never used" error.

peytondmurray:
I'm interested in seeing this change through. @asottile I think this is a case which isn't handled by `PYFLAKES_BUILTINS`, but if there's another way of getting around this issue I'd be interested in learning how.

jayvdb (approved):

li-dan:
Thanks!

PR #746, opened 2022-11-24
Remove submitting patches via mailing list

dannysepler:
This is a guess, but I doubt pyflakes wants to keep supporting this?


asottile (approved):
yep lol

PR #781, opened 2023-07-31
Add changelog URL to package metadata

intgr:
These links will be visible on the package's PyPI page:

> image

And also tools like Renovate bot use this to provide direct access to the changelog for a dependency update.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -43,6 +43,9 @@ setup(
     url="https://github.com/PyCQA/pyflakes",
     packages=["pyflakes", "pyflakes.scripts", "pyflakes.test"],
     python_requires='>=3.8',
+    project_urls={
+        "Changelog": "https://github.com/PyCQA/pyflakes/blob/main/NEWS.rst",
+    },
     classifiers=[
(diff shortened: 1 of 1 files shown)


asottile:
this has been declined multiple times -- please search next time.

intgr:
It didn't occur to me that this change might be controversial.

The only response I found, at https://github.com/PyCQA/pyflakes/pull/655#pullrequestreview-769506881, doesn't really address the motivation I laid out in the PR:

Did you consider that this is a quality of life improvement for users of Renovate bot?

When a pull request to update a package comes in, I want to see the changelog immediately. I don't want to spend time searching through the project's GitHub landing page, the README file or documentation for where a developer may have put a link to changelog, because every package does it differently.

Good tooling saves time and provides value for developers, like `pre-commit`.

asottile:
yes I want people to land on github and go from there

if renovate can't read the conventional location perhaps open an issue there.  but it is not our job to fix their bug

intgr:
There is no "conventional location". The convention is to add the URL to "project_urls".

Issue #779, opened 2023-07-30
Test regressions with PyPy3.10


mgorny:
The test suite fails with PyPy3.10 7.3.12:

```pytb
$ tox -e pypy310
pypy310: install_deps> python -I -m pip install flake8==6.0.0
.pkg: _optional_hooks> python /usr/lib/python3.11/site-packages/pyproject_api/_backend.py True setuptools.build_meta __legacy__
.pkg: get_requires_for_build_sdist> python /usr/lib/python3.11/site-packages/pyproject_api/_backend.py True setuptools.build_meta __legacy__
.pkg: get_requires_for_build_wheel> python /usr/lib/python3.11/site-packages/pyproject_api/_backend.py True setuptools.build_meta __legacy__
.pkg: prepare_metadata_for_build_wheel> python /usr/lib/python3.11/site-packages/pyproject_api/_backend.py True setuptools.build_meta __legacy__
.pkg: build_sdist> python /usr/lib/python3.11/site-packages/pyproject_api/_backend.py True setuptools.build_meta __legacy__
pypy310: install_package> python -I -m pip install --force-reinstall --no-deps /tmp/pyflakes/.tox/.tmp/package/3/pyflakes-3.1.0.tar.gz
pypy310: commands[0]> python --version --version
Python 3.10.12 (af44d0b8114cb82c40a07bb9ee9c1ca8a1b3688c, Jul 21 2023, 09:40:48)
[PyPy 7.3.12 with GCC 12.3.1 20230526]
pypy310: commands[1]> python -m unittest discover pyflakes
....F....F.F..................................................................s.................................s...............................................................................................s....................................................................................ss..s..........s....................s.........................................................................................................................................................................s.................................................................................s........s......................................................sssss..........s..................................................ss..s..........s.
======================================================================
FAIL: test_eofSyntaxError (test.test_api.CheckTests)
The error reported for source files which end prematurely causing a
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/tmp/pyflakes/pyflakes/test/test_api.py", line 450, in test_eofSyntaxError
    self.assertHasErrors(sourcePath, [expected])
  File "/tmp/pyflakes/pyflakes/test/test_api.py", line 338, in assertHasErrors
    self.assertEqual(
[...]

PR #802, opened 2024-02-14
Add py313-dev and pypy3.10 to CI

jayvdb:
fixes https://github.com/PyCQA/pyflakes/issues/779

Includes fixes to test_api.py to reflect the output of pypy3.10


asottile:
#779 was rejected:

> for now we only test one pypy version in CI and so we're just going to keep those positions there rather than trying to make the tests work for every version of pypy

Issue #813, opened 2024-06-25
Remove Python 2.x code?


edreamleo:
The latest readme says "[pyflakes] supports all active versions of Python: 3.8+".

Yet `checker.py` contains the following `counter` function:
```python
def counter(items):
    """
    Simplest required implementation of collections.Counter. Required as 2.6
    does not have Counter in collections.
    """
    results = {}
    for item in items:
        results[item] = results.get(item, 0) + 1
    return results
```
Why is this function necessary?

asottile:
it still works and nobody has bothered to remove it

PR #814, opened 2024-06-25
Replace 'counter' function with 'collection.Counter'

edreamleo:
See #813.


pyflakes/checker.py
```diff
@@ -1773,6 +1761,7 @@ def DICT(self, node):
         # Complain if there are duplicate keys with different values
         # If they have the same value it's not going to cause potentially
         # unexpected behaviour so we'll not complain.
+        counter = collections.Counter
```
asottile, line 1764: this should fix the call, this line looks suspicious as written

asottile (commented):
please remove "PR:" from your title and commit message.  of course it's a PR because it is

also please remove "all tests pass", we have ci checks in place which will be the judge of that

edreamleo:
@asottile Thanks for your guidance.

pyflakes/checker.py
edreamleo: @asottile rev 2d8f8de does as you requested.

edreamleo:
@asottile I would rather not change a previous commit message, but feel free to change it if think it's best.

sigmavirus24 (approved):

edreamleo:
@sigmavirus24 Thanks, Ian.

sigmavirus24:
Thanks for helping improve the project, @edreamleo

edreamleo:
@sigmavirus24 You're welcome :-)

Issue #812, opened 2024-06-16
`test_errors_syntax` fails with Python 3.13.0b2


mgorny:
The following test seems to fail due to improved error messages in Python 3.13:

```pytb
$ tox -e py313
py313: install_deps> python -I -m pip install flake8==6.1.0
.pkg: install_requires> python -I -m pip install 'setuptools>=40.8.0' wheel
.pkg: _optional_hooks> python /usr/lib/python3.12/site-packages/pyproject_api/_backend.py True setuptools.build_meta __legacy__
.pkg: get_requires_for_build_sdist> python /usr/lib/python3.12/site-packages/pyproject_api/_backend.py True setuptools.build_meta __legacy__
.pkg: get_requires_for_build_wheel> python /usr/lib/python3.12/site-packages/pyproject_api/_backend.py True setuptools.build_meta __legacy__
.pkg: install_requires_for_build_wheel> python -I -m pip install wheel
.pkg: prepare_metadata_for_build_wheel> python /usr/lib/python3.12/site-packages/pyproject_api/_backend.py True setuptools.build_meta __legacy__
.pkg: build_sdist> python /usr/lib/python3.12/site-packages/pyproject_api/_backend.py True setuptools.build_meta __legacy__
py313: install_package> python -I -m pip install --force-reinstall --no-deps /tmp/pyflakes/.tox/.tmp/package/1/pyflakes-3.2.0.tar.gz
py313: commands[0]> python --version --version
Python 3.13.0b2 (main, Jun 15 2024, 15:12:59) [GCC 14.1.1 20240516]
py313: commands[1]> python -m unittest discover pyflakes
..................F...........F...............................................s.................................s...............................................................................................s....................................................................................ss..s..........s....................s.........................................................................................................................................................................s............................................................................................s..........................................................................................................................ss..s..........s.
======================================================================
FAIL: test_errors_syntax (test.test_api.IntegrationTests.test_errors_syntax)
When pyflakes finds errors with the files it's given, (if they don't
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/tmp/pyflakes/pyflakes/test/test_api.py", line 779, in test_errors_syntax
    self.assertEqual(d, ('', error_msg, 1))
    ~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #821, opened 2024-11-19
FIx syntax assert error in tests

eamanu:
This commit add new syntax exception in assert only for Python3.13.

Closes: #812


sigmavirus24 (approved):

sigmavirus24:
Would you be willing to add 3.13 to the CI matrix?

asottile:
via #826

PR #822, opened 2025-01-23
Implement a possible fix for issue #137

hangie:
While I appreciate the issue was closed as not going to be fixed, I would like to raise this for consideration.

Happy to discuss further the merrits of doing or not doing this.

### DIFF
--- a/pyflakes/checker.py
+++ b/pyflakes/checker.py
@@ -1001,7 +1001,27 @@ class Checker:
                 # it may be a re-assignment to an already existing name
                 scope.setdefault(value.name, value)
             else:
-                self.scope[value.name] = value
+                if isinstance(value, SubmoduleImportation):
+                    # SubmoduleImportation is used for explicit imports
+                    # of submodule imports. Add the submodule full name to the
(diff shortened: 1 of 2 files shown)


asottile:
this has already been declined intentionally

hangie:
I get it was intentionally declined, however I can only see examples of why based on what was referred to as evil code that you shouldn't write in the first place. We are seeing failures downstream caused by the failure to pickup unused imports. I'm not tabling about static typing issues, I'm talking about runtime errors.

One option is to fix the identification of this kind of issue in the tools used to lint the Python code. Another is to write our own tool. I'm struggling to understand why the second is the recommended solution.

I believe you recommended pylint as to tool to use as you thought it would find this - it doesn't.

Would like to understand your suggestions

sigmavirus24:
Please read the thread closely: https://github.com/PyCQA/pyflakes/issues/137#issuecomment-519675359 is the philosophy of this project which is provided to you AS IS. You're not entitled to a bespoke re-explanation of something clearly explained.

PR #842, opened 2025-05-23
updates for python 3.14

asottile:


pyflakes/checker.py
```diff
@@ -1216,7 +1216,10 @@ def _enter_annotation(self, ann_type=AnnotationState.BARE):
     def _in_postponed_annotation(self):
         return (
             self._in_annotation == AnnotationState.STRING or
-            self.annotationsFutureEnabled
+            (
+                self._in_annotation == AnnotationState.BARE and
+                (self.annotationsFutureEnabled or sys.version_info >= (3, 14))
+            )
```
asottile, line 1222: there was a subtle bug here with the existing code -- I added a test for it below

sigmavirus24 (approved):

PR #844, opened 2025-08-14
fix typo

datdat:
I believe it should be 'without' side effects


asottile:
ah no the point is that even if the module has side effects it's still ok

PR #847, opened 2025-11-26
Add plugins folder. Add Checker.load_plugins and call it

edreamleo:
This PR adds support for pyflakes plugins, based loosely on the scheme used by pylint.


asottile:
flake8 is the intended plugin framework

edreamleo:
@asottile Thanks for the pointer.

PR #858, opened 2026-07-22
Report 'await' outside async functions

### LINKED ISSUE #818: Use of await in a function that is not async should be shown as a syntax error by flake8
Hello,

Reporting here, since flake8 seems to want this sort of problem delegated to you.

The following code uses await in a function that is declared without the async keyword, which is a syntax error.
```
def foo():
    pass


def bar():
    await foo()
```

Running flake8 on this succeeds, no error is reported.

flake8 --version
7.1.1 (mccabe: 0.7.0, pycodestyle: 2.12.1, pyflakes: 3.2.0) CPython 3.11.2 on
Linux
Debian 12.7
pyflakes 3.2.0 installed in a virtual environment

l46983284-cpu:
`await` is only valid inside `async def`. pyflakes already warns for `yield` / `return` outside functions, but `AWAIT` was aliased to the yield handler, so `await` inside a plain `def` produced no message.

Track async function scopes and report `await` outside them.

Fixes #818

### DIFF
--- a/pyflakes/checker.py
+++ b/pyflakes/checker.py
@@ -824,8 +824,8 @@ class Checker:
         return self.scopeStack[-1]
 
     @contextlib.contextmanager
-    def in_scope(self, cls):
-        self.scopeStack.append(cls())
+    def in_scope(self, cls, *args, **kwargs):
+        self.scopeStack.append(cls(*args, **kwargs))
         try:
             yield
         finally:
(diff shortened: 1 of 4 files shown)


pyflakes/checker.py
```diff
@@ -1934,14 +1936,30 @@ def YIELD(self, node):
...
+
+        self.handleNode(node.value, node)
 
     def FUNCTIONDEF(self, node):
         for deco in node.decorator_list:
             self.handleNode(deco, node)
 
         with self._type_param_scope(node):
-            self.LAMBDA(node)
+            self.LAMBDA(node, is_async=isinstance(node, ast.AsyncFunctionDef))
```
sigmavirus24: Lambdas can't be async so why do we need this?

sigmavirus24 (changes requested):

alexchen-sys:
Thanks — you're right, lambdas aren't async. I dropped the is_async parameter from LAMBDA and derive the flag only from AsyncFunctionDef when the shared body helper builds FunctionScope, so the await-outside-async check still works without implying async lambdas.

